PR #34, opened 2015-06-25
Active queue sort

teonar:
Easy overview of active jobs. Controlled by checking/unchecking
a textbox.  If checked will show all jobs with status not queued first.

### DIFF
--- a/webui/downloads.js
+++ b/webui/downloads.js
@@ -17,8 +17,8 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * $Revision$
- * $Date$
+ * $Revision: 1236 $
+ * $Date: 2015-03-20 16:02:44 -0400 (Fri, 20 Mar 2015) $
  *
  */
 
@@ -36,6 +36,7 @@ var Downloads = (new function($)
 
 	// Controls
 	var $DownloadsTable;
+	var $DownloadsShow;
 	var $DownloadsTabBadge;
 	var $DownloadsTabBadgeEmpty;
 	var $DownloadQueueEmpty;
@@ -43,6 +44,7 @@ var Downloads = (new function($)
 	var $DownloadsTable_Name;
 
 	// State
+	var showActive = false;
 	var notification = null;
 	var updateTabInfo;
 	var groups;
@@ -66,6 +68,7 @@ var Downloads = (new function($)
 		'EXECUTING_SCRIPT': { Text: 'PROCESSING', PostProcess: true },
 		'PP_FINISHED': { Text: 'FINISHED', PostProcess: false }
 		};
+
 	this.statusData = statusData;
 
 	this.init = function(options)
@@ -79,8 +82,23 @@ var Downloads = (new function($)
 		$DownloadsRecordsPerPage = $('#DownloadsRecordsPerPage');
 		$DownloadsTable_Name = $('#DownloadsTable_Name');
 
+
+
 		var recordsPerPage = UISettings.read('DownloadsRecordsPerPage', 10);
 		$DownloadsRecordsPerPage.val(recordsPerPage);
+
+		showActive = UISettings.read('DownloadsShowActive', false);
+		// looks like the write saves as string, so translate it back
+		if (showActive == 'true')	
+			showActive = true;
+		else
+			showActive = false;
+
+// no idea why this doesn't work
+//		var DownloadsShowActive = $('#DownloadsShowActive');
+//		DownloadsShowActive.checked = showActive;
+// and this does
+		$("#DownloadsShowActive").prop("checked", showActive);
 
 		$DownloadsTable.fasttable(
 			{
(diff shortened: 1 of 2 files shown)


hugbug:
Thank you for your changes.
I've tried them but found few issues:
1. Hitting the play/pause button changes the order of items in the queue. This is understandable but confusing.
2. If you change priority of an item the queue may change as well and you may not see the item on the same page anymore.
3. The move-buttons work weird since the items are moved in their original positions, not in positions they are shown. The effect of clicking move-buttons may not be seen immediately and the items don't respond to move-actions as expected.
4. A better UI for this feature is needed; may be a button on the toolbar.  I've added a screenshot of current implementation as reference for other users.
   screen shot 2015-06-26 at 00 25 44

On the whole I think It's not a good idea to show items in wrong order.

Can we take a step back and discuss why the change was necessary? You haven't explained this but I think it's because you have many items in queue with different priorities. If you would have all items with one priority you wouldn't have the problem at all - the items will be processed in order they are displayed.

We currently have **on-demand queue sorting** added in v15 - forum topic [[New Feature] On-demand queue sorting](http://nzbget.net/forum/viewtopic.php?f=10&t=1789).

Column **Status** isn't clickable yet. What if we make it clickable? That would reorder the items in a similar order like you did. The difference is the order is permanent and this solves all mentioned problems. A disadvantage is - once another item starts downloading you may need to click it again.

What do you think?

teonar:
I do have a large queue with different priorities. Using the same priority for everything will not work as I want some items to download before others (highest priority downloads first)

Making the <b>Status</b> sortable would solve the problem only if the list is sorted on each iterface refresh. Having to interact with the interface to monitor is wrong in my opinion.

Another option is to have the list sort on <b>Priority</b>.

hugbug:
> I do have a large queue with different priorities. Using the same priority for everything will not work as I want some items to download before others.

You shouldn't use the same priority of course. That would negate the purpose of priorities.

> Making the Status sortable would solve the problem only if the list is sorted on each iterface refresh. Having to interact with the interface to monitor is wrong in my opinion.

I meant only sorting on demand (with click).

**Then we need another solution.**

teonar:
Think that sorting on priority would be best but I think it may place extra stress on the server.

Or maybe just a tab 
e.g. nzbget1
Then in the <b>Active</b> tab display everything with a non-queued status.

hugbug:
I agree that sorting on priority is probably the best solution.

Have you used SABnzbdb+ before? Because that's how it works there:
- all items are sorted by priority;
- when a new item is added it is put to queue according to it's priority (the one could say the queue is resorted after a new nzb is added);
- if you change priority of an item it is automatically moved in the queue after the items with the same priority (the one could say the queue is resorted after items are edited);
- you can move items only inside their priority window (the one could say the queue is resorted after items are moved on user's request).

I didn't like this before but I also understand the flaws of current behavior. If making it like in SAB I would add an option (name **PrioritySort=yes/no** comes to mind) to allow users to choose the preferred behavior.

The internal function to sort was added in v15. It can sort by priority too. What we need is to find the places in the code where the items are added or moved and call the sorting routing after that. Since this will happen only after certain actions I don't expect any significant load on server caused by that feature.

Would you like to implement that? That would be a C++ part, I could give tips about where to look.

teonar:
Yes I know how sabnzbd displays its entries. That would work but still not really what I want. If it is sorted by priority then you lose the visual display of the order that the items were added.

How about the following idea.

Checkbox that acts like a predefined filter. 
When checked it filters and displays only non-queued items (sets the downloads badge to yellow to indicate filter results and updates the text at the end of the table)
When unchecked the full list is displayed.

Think this would be more in line with the nzbget design..

teonar:
On final thought, what I really want is a <b>not</b> and <b>or</b> operator in the search 

Can already search for the following, just cannot search for a subset.
QUEUED
FETCHING
DOWNLOADING
PP_QUEUED
PAUSED
LOADING_PARS
VERIFYING_SOURCES
REPAIRING
VERIFYING_REPAIRED
RENAMING
MOVING
UNPACKING
EXECUTING_SCRIPT
PP_FINISHED

Currently the search breaks up the string into words, all words must be in the search string.

hugbug:
I feel we are going in the right direction :thumbsup:
1. We can do it similar to filters in tabs **history** and **messages**:
   
   screen shot 2015-06-26 at 14 08 49
   with **All** and **Active**
   
   There is no much place for new buttons on the toolbar, but that depends on resolution of course.
2. Or predefined filters like GitHub has on issues page:
   
   screen shot 2015-06-26 at 14 16 04

Advanced syntax for search string would be a fantastic feature I always missed but never had time to work on it. That would be a great addition regardless of whether the alternative 1 or 2 will be implemented. All search fields would profit from it.

We should keep things consistent. Making filters via search field/combo on downloads tab but leaving the history and messages tabs as is would be inconsistent. I'm currently prefer the first alternative because it's easier for usage (I think).

On another though, having badges for active/all isn't very helpful because there is usually only one active item anyway. Instead of two buttons [ All (10) | Active (1) ] we can do that only via one button [ Active ] similar to button [ Hidden ] on history page which works like a checkbox:

screen shot 2015-06-26 at 14 35 36

It also takes less place than two buttons with badges.

---

I wish more input from other users. Calling [stargazers](https://github.com/nzbget/nzbget/stargazers) @prinz23, @rcdailey and @JVMed.

teonar:
> Advanced syntax for search string would be a fantastic feature I always missed but never had time to >work on it. That would be a great addition regardless of whether the alternative 1 or 2 will be >implemented. All search fields would profit from it.

I implemented it with (!) and (|) in the search text. Now just need to figure out how to merge/push it, having some trouble with git :p

PR #36, opened 2015-06-26
Implemented or(|) not(!) qualifiers.

teonar:
e.g.
!earth !moon
|earth |moon
!earth moon
!earth |moon


teonar:
Wrong branch

PR #37, opened 2015-06-26
Implemented or(|) not(!) qualifiers. and(&) is the default.

teonar:
e.g.
!earth !moon
|earth |moon
!earth moon
!earth |moon

### DIFF
--- a/webui/fasttable.js
+++ b/webui/fasttable.js
@@ -17,8 +17,8 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * $Revision: 778 $
- * $Date: 2013-08-07 16:09:43 -0400 (Wed, 07 Aug 2013) $
+ * $Revision$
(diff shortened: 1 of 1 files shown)


webui/fasttable.js
```diff
@@ -17,8 +17,8 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * $Revision$
- * $Date$
+ * $Revision: 778 $
+ * $Date: 2013-08-07 16:09:43 -0400 (Wed, 07 Aug 2013) $
```
hugbug: Please revert these two lines. Then I can merge the pull request without editing.

Looks like you were using the files from svn as the basis for the editing. The files from git doesn't have the revision and date as git doesn't support keywords (unfortunately).

teonar: aah that is why I was having problems!

will revert.

teonar: Apologies, worked from other sources that was in a gz file.

PR #38, opened 2015-06-26
Implemented or(|) not(!) qualifiers. and(&) is the default.

teonar:
e.g.
!earth !moon
|earth |moon
!earth moon
!earth |moon

Hopefully this time I got it right.


hugbug:
For the future: It wasn't necessary to create a new pull request. Your new commits are automatically added to the existing request.

teonar:
Ok so now that the search terms are in, How abouot adding something like in
the attachment ?

ps, github wont allow me to upload the image.

PR #40, opened 2015-06-27
Predefined search terms

teonar:
Suggestion for predefined search terms.

Todo:
search icon needed (used the config icon for demonstration)
letting user define terms in config file (do not know if this is needed/wanted)

PP-Queued should also be added to active list.
Boolean implementation was improved.

### DIFF
--- a/webui/downloads.js
+++ b/webui/downloads.js
@@ -41,6 +41,7 @@ var Downloads = (new function($)
 	var $DownloadQueueEmpty;
 	var $DownloadsRecordsPerPage;
 	var $DownloadsTable_Name;
+	var $DownloadsTable_Filter;
 
 	// State
 	var notification = null;
@@ -78,6 +79,8 @@ var Downloads = (new function($)
 		$DownloadQueueEmpty = $('#DownloadQueueEmpty');
 		$DownloadsRecordsPerPage = $('#DownloadsRecordsPerPage');
 		$DownloadsTable_Name = $('#DownloadsTable_Name');
+		$DownloadsTable_Filter = $('#DownloadsTable_filter'); 
+
 
 		var recordsPerPage = UISettings.read('DownloadsRecordsPerPage', 10);
 		$DownloadsRecordsPerPage.val(recordsPerPage);
@@ -554,6 +557,12 @@ var Downloads = (new function($)
 		notification = '#Notif_Downloads_Sorted';
 		RPC.call('editqueue', ['GroupSort', 0, order, checkedEditIDs], editCompleted);
 	}
+
+	this.filter = function(type)
+	{
+		$DownloadsTable_Filter.val(type);
+	}
+
 }(jQuery));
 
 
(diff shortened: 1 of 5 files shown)


teonar:
PP-Queued should also be added to active list.
Boolean implementation was improved.

hugbug:
Instead of the new button the menu should be integrated into the search box:

screen shot 2015-06-27 at 19 29 00

Doable?

teonar:
I think I can do it, just did not think of it when playing around.

hugbug:
Great,

BTW, the current trend is to use rectangular boxes (instead of rounded) for search boxes. If that helps to put the menu button there you may do that change too. Only with very small round radius, like the toolbar buttons.

This search box is used on all tabs (downloads, history, messages, settings). Don't write code to hide the menu button on other tabs, we'll make it useful on all tabs later.

teonar:
Something like this, obviously still work to to, just look at the idea.

history/message/settings is wrong, think the controls get hidden by the fasttable class, have not looked too closely yet.

teonar:
Some interesting things that users probably do not know (seeing it in the code)
e.g. pressing <b>Enter</b> make the search case sensitive

BTW this search terms only orks if refresh is active, have not looked at how to immediately refresh like the <b>Clear</b> button

hugbug:
Good progress here :thumbsup:

Suggestions:
- the menu should not be shown when the edit box is clicked. The menus typically shown in such cases are for autocomplete, which isn't our case;
- therefore we do need the arrow (caret) icon which will trigger the menu; Once again the GitHub's issue search box as an example (the "Filters"-menu):
  
  screen shot 2015-06-28 at 18 16 57
  
  We don't have that much space to add a text, a small icons is enough.
- current implementation doesn't work on Firefox - a click on the search box opens the menu but doesn't set focus to the edit box, the search box isn't usable therefore. I think this problem will be automatically fixed when the previous issue is solved.

For the future improvements: instead of hard coding the search terms I would implement an ability for users to alter the menu. The last three menu items would be: _save search_, _remove search_ and _remove all_; These are the only menu items the user sees when using the menu the first time. The search menu should be saved per tab (download, history, etc.). This is however the second todo. First is to finish the UI - hard coded items are OK for that phase.

Another note: the UI has two modes: **desktop** and **phone**. If you make the browser window small enough you'll see how the phone mode activates. In this mode many UI elements are shown differently. Most of this is done via extra CSS formatting. Sometimes javascript is involved too. When doing changes to UI don't forget to checked the phone mode too.

Thank you for your contribution.

hugbug:
Looks like you've the lost the interest for this feature.

After implementing of advanced search in #72 the predefined search terms would be even more useful.

This PR however conflicts with current develop branch and it also contains experimental code (different versions) which is bad for commit history.

If you still interested in this feature please feel free to open a clean PR based on current develop branch. Before working on it please let me know about that because I may start implementing this myself as explained in my previous post - https://github.com/nzbget/nzbget/pull/40#issuecomment-116295441.

teonar:
Hi Andrey
Have not lost interest, Had a burglary at my home and my stuff was taken, still trying to get everything back to what it was.No idea when I will be able to do anything, probably 3 weeks or more.
RegardsT

PR #42, opened 2015-06-29
Show correct LAN/WLAN ip address after installing

sanderjo:
As discussed on http://nzbget.net/forum/viewtopic.php?f=3&t=2030

### DIFF
--- a/linux/installer.sh
+++ b/linux/installer.sh
@@ -482,15 +482,19 @@ if test "$JUSTUNPACK" = "no"; then
         Info "Successfully installed into $ABSOUTDIR"
 
         # Trying to get current IP-address        
-        IP=""
-        {
-            IP=`ifconfig | sed -rn 's/.*r:([^ ]+) .*/\1/p' | head -n 1` || true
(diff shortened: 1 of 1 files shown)


hugbug:
Thank you.

The tricky part about installer is that it should work on many systems. It was tested on many busybox variants. With busybox you never know what commands are available and how much they were stripped. 

Since we can't expect such advanced commands like "ifconfig" to work we must wrap the calls with safety straps like `{ .. } > /dev/null 2>&1` - suppress error reporting, for example if "ifconfig" isn't available.

To test your code for durability replace "route" with "route1" and "ifconfig" with "ifconfig1" to pretend the commands are not available.

Don't worry, I'll add the safety wrapper myself. Thanks again for the improvement.

hugbug:
- removed the info message for local address. The info about web-interface were supposed to be a one-liner, not three lines. It's primary for users who install on remote machines. Local address 127.0.0.1 works always anyway. 
- good idea about printing default login name and password :+1: 
- added safety wrapper `{..}`.

PR #129, opened 2015-12-19
Added trim method to url string value

fjeldsoe:
Adding a NZB URL fails if there is a space at the start of the string, this should fix it. Often when you copy and paste a link, spaces in front of the string can occur.


hugbug:
Thanks!

Issue #125, opened 2015-12-10
Activate C++14 compiler features


hugbug:
For motivation see #88.
This is related to #101.

To use new C++ features we have already upgraded to newer compilers. However in order to activate these new features current compilers requires extra parameters.

In this issue:
- [x] update Linux build script to activate C++14 when compiling NZBGet;
- [x] update documentation for those who compile on their own with hints on how to activate C++14 in GCC and Clang;
- [x] extend configure script to check for required compiler features.

hugbug:
Added compiler features check. Compiling now works without explicit setting of compiler flags, with usual

``` shell
./configure
make
```

It should automatically add required compiler switches.

It also reports an error if compiler is too old. The check can be disabled with `./configure --disable-cpp-check`, in which case required compiler options must be set via cpp-flags manually.

PR #134, opened 2015-12-26
Specify -std=c++11 option in Makefile.am

rcdailey:
Compilation was failing due to the lack of this option. C++11 features are currently in use, requiring this compiler flag.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -196,6 +196,7 @@ nzbget_SOURCES += \
 endif
 
 AM_CPPFLAGS = \
+	-std=c++11 \
 	-I$(srcdir)/daemon/connect \
 	-I$(srcdir)/daemon/extension \
 	-I$(srcdir)/daemon/feed \


hugbug:
Thanks.

This will be addressed as part of #125.
Adding compiler specific flags into automake-file isn't a good idea, I think.

For gcc this option is enough but for example clang (at least on Mac) requires additional option: `-std=c++14 -stdlib=libc++`

I'll update configure script with detection for C++14, which will also print a link to nzbget installation wiki if C++14 wasn't found or was not active.

This is how I do that:

```
CXXFLAGS="-std=c++14" ./configure
```

rcdailey:
I'm not familiar with linux development, especially make/automake/etc. This is the change I made just poking around, and it seemed to get things compiling for me. I leave the best solution to you, then. However note that latest on `develop` does not compile with GCC 5.2 on Ubuntu 15.10. I do see the changes you made to introduce C++14, however those flags are not being passed to GCC when I build. The commands I run (always) are:

```
$ git pull
$ ./configure
$ make -j4
```

These are per the instructions found in your `README` files. If there are extra steps I'm missing, let me know.

I also find it weird that you need to explicitly specify `-stdlib` on mac.

hugbug:
> does not compile with GCC 5.2 on Ubuntu 15.10

GCC 5.2 is definitely fine. I use it too in a Linux VM.

Please try this:

```
CXXFLAGS="-std=c++14" ./configure
```

If it fails during configure please post `config.log` (via pastern.com). If it fails during compiling post the compilation log (copy from terminal output or redirect compiler to a file).

rcdailey:
With the flags set via environment, of course it will work. However this is undocumented and IMHO an ad-hoc fix. My point was that via normal instructions, it was not working. We already have both established that the proper, long-term fix is to have the build system be aware of what level of the C++ standard is required.

Thanks for clarifying, I'll close this PR.

hugbug:
Found an autoconf-macro to test for C++14 and automatically add required flags - [AX_CXX_COMPILE_STDCXX_14](https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_14.html#ax_cxx_compile_stdcxx_14).

I have to integrate it into configure-script; will probably require an upgrade of configure-script to a newer autotools version first.

rcdailey:
I don't mean to go off on some distant tangent here, but it might be worthwhile to consider a different build system, such as [CMake](https://cmake.org/). I've used it for over 10 years and it's the most functional, portable, and modern build system to date.

I personally don't understand why people deal with linux make tools like this. You've probably already given it some thought but I wanted to voice my opinion. If you are considering it, I can help you setup some CMake scripts.

Prinz23:
Just wanted to add: Debian Jessie (current stable) is still on gcc 4.9 
So people with Debian have a problem now: nzbget doesn't compile

rcdailey:
@Prinz23 Not sure what you're saying? GCC 4.9 supports `-std=c++14`.

hugbug:
GCC declares full support of C++14 in version 5.1.
Older versions support some features but not all.

I'm not going to use every new feature; at the moment I use only a few, most notably the move semantics, which is a core for some refactorings in NZBGet.

Just tested with GCC **4.8** and it worked:

```
CXXFLAGS="-std=c++11" ./configure
```

From [this feature list](https://gcc.gnu.org/projects/cxx0x.html) I think it will even work with GCC 4.6.

Prinz23:
Yeah I wanted mainly to remind you that some systems don't have full featured C++14 support, since this a work in progress and I don't know what will be used or not.

With the flag it will currently work, as long as no features are used that aren't supported.

hugbug:
Thanks for reminding. I wasn't aware about Debian being on 4.9 still. Do you know when Debian stable become gcc 5?
Not that I need some specific feature of 5.x right now; it's just more comfortable to know that anything I read in new books can be used.

Prinz23:
I was planned to be added last summer, but too many issues with it. Last that I read said: It will probably not be added to Jessie stable, but it's successor Stretch. But depends on if the issues can be resolved and when.

hugbug:
Added compiler features check. Please try compiling the latest **develop** with usual

``` shell
./configure
make
```

It should automatically add required compiler switches.

It also reports an error if compiler is too old. The check can be disabled with `./configure --disable-cpp-check`, in which case required compiler options must be set via cpp-flags manually.

PR #189, opened 2016-03-20
Add the ax_cxx_compile_stdcxx macro locally

sbraz:
It's probably best to add this macro locally instead of relying on the system to provide it. Especially since it's rather new and only present in the latest autoconf-archive release.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,7 @@
 
 AC_PREREQ(2.59)
 AC_INIT(nzbget, 17.0-testing, hugbug@users.sourceforge.net)
+AC_CONFIG_MACRO_DIR([posix])
 AC_CONFIG_AUX_DIR(posix)
 AC_CANONICAL_TARGET
 AM_INIT_AUTOMAKE([foreign])
--- a/posix/ax_cxx_compile_stdcxx.m4
+++ b/posix/ax_cxx_compile_stdcxx.m4
@@ -1,0 +1,565 @@
+# ===========================================================================
+#   http://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_CXX_COMPILE_STDCXX(VERSION, [ext|noext], [mandatory|optional])
+#
+# DESCRIPTION
+#
+#   Check for baseline language coverage in the compiler for the specified
+#   version of the C++ standard.  If necessary, add switches to CXXFLAGS to
+#   enable support.  VERSION may be '11' (for the C++11 standard) or '14'
+#   (for the C++14 standard).
+#
+#   The second argument, if specified, indicates whether you insist on an
+#   extended mode (e.g. -std=gnu++11) or a strict conformance mode (e.g.
+#   -std=c++11).  If neither is specified, you get whatever works, with
+#   preference for an extended mode.
+#
+#   The third argument, if specified 'mandatory' or if left unspecified,
+#   indicates that baseline support for the specified C++ standard is
+#   required and that the macro should error out if no mode with that
+#   support is found.  If specified 'optional', then configuration proceeds
+#   regardless, after defining HAVE_CXX${VERSION} if and only if a
+#   supporting mode is found.
+#
+# LICENSE
+#
+#   Copyright (c) 2008 Benjamin Kosnik <bkoz@redhat.com>
+#   Copyright (c) 2012 Zack Weinberg <zackw@panix.com>
+#   Copyright (c) 2013 Roy Stogner <roystgnr@ices.utexas.edu>
+#   Copyright (c) 2014, 2015 Google Inc.; contributed by Alexey Sokolov <sokolov@google.com>
+#   Copyright (c) 2015 Paul Norman <penorman@mac.com>
+#   Copyright (c) 2015 Moritz Klammler <moritz@klammler.eu>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved.  This file is offered as-is, without any
+#   warranty.
+
+#serial 1
+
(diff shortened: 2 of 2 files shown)


hugbug:
> instead of relying on the system to provide it

What do you mean? M4-files are not needed to build nzbget. Only developers making changes to configure.ac may need to execute "aclocal".

I don't like adding external files to the project directory tree unless this is really necessary.

I'm not an autotools expert though and may be doing things wrong.

sbraz:
Gentoo, for instance, always runs `autoreconf` to build the package. If we don't add the macro, the build will fail.
We currently add the file via a patch but I think it would be easier if the upstream tarball provided it. I'm not an autotools expert myself but I was told that this is a commonly accepted practice.
A bit of research showed that [gparted](https://github.com/GNOME/gparted/commit/f97cac9ab459bd2c2b8cd1098f7683a5ec137c50), [varnish](https://github.com/varnishcache/varnish-cache/tree/master/m4), [bitcoin](https://github.com/libbitcoin/libbitcoin/tree/master/m4), [kodi](https://github.com/xbmc/xbmc/tree/master/m4), [znc](https://github.com/znc/znc/tree/master/m4) and probably a lot of other projects do this.

sbraz:
Sorry, I meant [Gentoo's ebuild for nzbget runs autoreconf](https://github.com/gentoo/gentoo/blob/master/net-nntp/nzbget/nzbget-16.4.ebuild#L54), not all packages do.

sbraz:
Of course, none of this is critical :)
Once the release includes the fixed configure with the improved ncurses/tinfo detection, we will probably be able to skip the call to autoreconf.
I just thought it would be a nice addition for people who might try to run autoreconf and would then see configure fail with a rather cryptic error:

```
./configure: line 3914: syntax error near unexpected token `14,,optional'
./configure: line 3914: `       AX_CXX_COMPILE_STDCXX(14,,optional)'
```

hugbug:
Can we at least put the file into existing directory "posix" instead of creating a new one? I don't know if aclocal is OK with extra unrelated files in the directory.

sbraz:
Re-pushed. It seems to work fine this way too.

hugbug:
Great, thank you.
One last thing. You are using the default version of the macro. I'm using however a modified one, which works fine with gcc 4.9 (although it doesn't fully support C++14, it's OK for nzbget nonetheless).
Can you please extract the macro from [aclocal.m4](https://github.com/nzbget/nzbget/blob/develop/aclocal.m4#L1009)?

sbraz:
Re-pushed with your modified macro. This is the diff to the vanilla one, let me know if I did that properly:

``` diff
--- ax_cxx_compile_stdcxx.m4.1  2016-03-20 18:16:40.176167776 +0100
+++ posix/ax_cxx_compile_stdcxx.m4  2016-03-20 18:13:16.712371158 +0100
@@ -68,9 +68,12 @@
     ac_success=yes
   fi

-  m4_if([$2], [noext], [], [dnl
+  m4_if([$2], [ext], [], [dnl
   if test x$ac_success = xno; then
-    for switch in -std=gnu++$1 -std=gnu++0x; do
+    dnl HP's aCC needs +std=c++11 according to:
+    dnl http://h21007.www2.hp.com/portal/download/files/unprot/aCxx/PDF_Release_Notes/769149-001.pdf
+    dnl Cray's crayCC needs "-h std=c++11"
+    for switch in -std=c++$1 -std=c++0x +std=c++$1 "-h std=c++$1"; do
       cachevar=AS_TR_SH([ax_cv_cxx_compile_cxx$1_$switch])
       AC_CACHE_CHECK(whether $CXX supports C++$1 features with $switch,
                      $cachevar,
@@ -88,12 +91,9 @@
     done
   fi])

-  m4_if([$2], [ext], [], [dnl
+  m4_if([$2], [noext], [], [dnl
   if test x$ac_success = xno; then
-    dnl HP's aCC needs +std=c++11 according to:
-    dnl http://h21007.www2.hp.com/portal/download/files/unprot/aCxx/PDF_Release_Notes/769149-001.pdf
-    dnl Cray's crayCC needs "-h std=c++11"
-    for switch in -std=c++$1 -std=c++0x +std=c++$1 "-h std=c++$1"; do
+    for switch in -std=gnu++$1 -std=gnu++0x; do
       cachevar=AS_TR_SH([ax_cv_cxx_compile_cxx$1_$switch])
       AC_CACHE_CHECK(whether $CXX supports C++$1 features with $switch,
                      $cachevar,
@@ -110,6 +110,7 @@
       fi
     done
   fi])
+
   AC_LANG_POP([C++])
   if test x$ax_cxx_compile_cxx$1_required = xtrue; then
     if test x$ac_success = xno; then
@@ -445,7 +446,11 @@

 #error "This is not a C++ compiler"

-#elif __cplusplus < 201402L
+//check for full C++14 support (disabed)
+//#elif __cplusplus < 201402L
+
+// check for partial C++14 support (accept gcc 4.9)
+#elif __cplusplus < 201300L

 #error "This is not a C++14 compiler"

@@ -480,6 +485,7 @@

   }

+/*
   namespace test_generalized_constexpr
   {

@@ -499,6 +505,7 @@
     static_assert(strlen_c("another\0test") == 7UL, "");

   }
+*/

   namespace test_lambda_init_capture
   {
```

hugbug:
For some reason in my test the macro wasn't integrated into `configure` and configure failed then with an error.

What did help was adding the following line to `configure.ac` (as suggested [here](https://autotools.io/autoconf/macros.html)):

```
m4_include([posix/ax_cxx_compile_stdcxx.m4])
```

I've committed this into branch [pr/189](https://github.com/nzbget/nzbget/tree/pr/189). Can you please test if this version works for you (before I merge into develop)?

sbraz:
Yes, this branch generates a working configure and compiles fine. Thanks for looking into this!

hugbug:
Great. Thank you for the contribution.

PR #190, opened 2016-03-20
Removed in-class member initializers for VolumeArray sizes

heksesang:
The code `VolumeArray m_bytesPerSeconds{60};` actually initializes a vector of size 1, with 60 as the first element. This caused out of bounds exceptions once it tried to write to higher indices than 0 at line 92 in StatMeter.cpp.

To initialize it to size 60, I have added a constructor that does the initialization.

### DIFF
--- a/daemon/nntp/StatMeter.h
+++ b/daemon/nntp/StatMeter.h
@@ -55,9 +55,9 @@ public:
 	void LogDebugInfo();
 
 private:
-	VolumeArray m_bytesPerSeconds{60};
-	VolumeArray m_bytesPerMinutes{60};
-	VolumeArray m_bytesPerHours{24};
+	VolumeArray m_bytesPerSeconds = VolumeArray(60);
+	VolumeArray m_bytesPerMinutes = VolumeArray(60);
+	VolumeArray m_bytesPerHours = VolumeArray(24);
 	VolumeArray m_bytesPerDays;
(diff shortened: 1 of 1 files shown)


hugbug:
> actually initializes a vector of size 1, with 60 as the first element. 

:scream:

The following seems to do the job and is the shortest:

``` C++
    VolumeArray m_bytesPerSeconds = VolumeArray(60);
    VolumeArray m_bytesPerMinutes = VolumeArray(60);
    VolumeArray m_bytesPerHours = VolumeArray(24);
```

If you have nothing against this can you please update the PR to utilize this version? Can you also make the PR to consist of only one commit (rebase or something) if that's possible (to avoid growing of history with small fix-commits)?

Thank you for that fix :+1: .

heksesang:
Sure, I'll do that. I do agree that looks better. :)

heksesang:
Rebased and ready for pull. :)

hugbug:
Thanks a lot. That was a serious bug.

hugbug:
If you take a look at [commits](https://github.com/nzbget/nzbget/commits/develop) you'll see that your commit isn't linked to your user page and consequently you don't appear in contributor list of nzbget project.

This is probably caused by different e-mail-addresses configured in git (on a machine you created the commit) and in your github account. You might need to check this.

Issue #88, opened 2015-10-11
Modern C++


hugbug:
### Intro

When developing NZBGet one of the priorities always was to make it work on as many devices as possible. Due to a wide variety of platforms we could not control which version of compiler is used. To make it compatible even with old compilers NZBGet was targeting old C++ versions.

In last years C++ has been evolved dramatically. New versions were standardized by ISO as C++11 and lately C++14. These new standards bring many interesting features to the language making it much more pleasure to work with.

In the meantime major Linux distributions have updated their compilers to GCC 5, which supports C++14. Xcode on OS X includes C++14 compatible compiler. MS Visual Studio 2015 is compatible with C++14 as well.

Moreover after the introducing of universal Linux installer in v15 we now offer precompiled binaries for three major platforms (Windows, Linux, OS X). This makes it less important which compiler is installed on user machine since only few users compile themselves.

So now it is the time to upgrade the code base of NZBGet to C++14.
### First steps (prerequisites)
- [x] upgrade Linux build machine to use GCC 5.2 for all targets included in universal installer (#101);
- [x] upgrade Window build machine to use MS Visual Studio 2015 (#102).
### Deep refactoring and new C++ features
- [x] change source code naming convention (#103);
- [x] put all external headers together (#115);
- [x] improve code formatting (#176);
- [x] use precompiled headers (#115);
- [x] use size specific integer types (#116);
- [x] use string classes for managed string handling (#126);
- [x] use namespaces (#133);
- [x] `nullptr` instead of `NULL` (#138);
- [x] store objects directly in containers (#143);
- [x] avoid "malloc/free" for temporary buffers (#156);
- [x] for-range loop with iterators (#152);
- [x] use lambdas with containers (#154);
- [x] prefer references to pointers (#151);
- [x] prefer stack objects (#167);
- [x] prefer safe memory management alternatives to plain "malloc/free" (#172);
- [x] use in-class member initializers (#175);
- [x] use std::shared_ptr for shared resources (#182);
- [x] indicate object ownership with std::unique_ptr (#148, #168);
- [x] exception safe (RAII) lock handling (#185).
### New standard libraries
-  ~~built-in **regex**-library instead of external regex~~ - _optional_;
-  ~~built-in **concurrency**-library instead of system specific pthreads or windows threads~~ - _optional_.

hugbug:
### GCC versions

[...]

PR #193, opened 2016-03-24
Fixed guards.

heksesang:
The way DownloadQueue::Guard was used in foreach loops, it would destroy the guard before iteration started (as the object wasn't assigned to any variable). This caused exceptions when the file list was modified during iteration, which could happen when the lock wasn't in place.

Also, I replaced the Mutex class with std::mutex, to further the rewrite to C++14.

### DIFF
--- a/daemon/frontend/NCursesFrontend.cpp
+++ b/daemon/frontend/NCursesFrontend.cpp
@@ -338,8 +338,8 @@ int NCursesFrontend::CalcQueueSize()
 	}
 	else
 	{
-        GuardedDownloadQueue guard = DownloadQueue::Guard();
-		for (NzbInfo* nzbInfo : guard->GetQueue())
+		GuardedDownloadQueue downloadQueue = DownloadQueue::Guard();
+		for (NzbInfo* nzbInfo : downloadQueue->GetQueue())
 		{
 			queueSize += nzbInfo->GetFileList()->size();
 		}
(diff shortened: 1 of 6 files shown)


heksesang:
Just a bit more explanation of the problem:

```
for (NzbInfo* nzbInfo : DownloadQueue::Guard()->GetQueue())
{
    ...
}
```

In the above code, the object protecting the queue, created by DownloadQueue::Guard(), is destroyed after the GetQueue() call is done returning, because the call chain is complete at that point.

Calls like this are fine, because the calls are chained:
`DownloadQueue::Guard()->GetQueue()->size()`

hugbug:
### Thank you

Another serious bug spotted. Thank you so much.

I was under impression the temporary object (guard) will live the whole loop life but I was wrong:

> If range_expression returns a temporary, its lifetime is extended until the end of the loop, as indicated by binding to the rvalue reference __range, but beware that the lifetime of any temporary within range_expression is not extended.

### Regarding `std::mutex`

I've [decided](https://github.com/nzbget/nzbget/issues/88#issuecomment-198696779) to not use standard libraries for **regex** and **concurrency** for now. That's to make sure the new bugs are caused by internal changes in NZBGet.

I may upgrade to `std::mutex` and `std::thread` in the future.

### One stylistic request

I'm using temporary name **downloadQueue** if the guarded pointer is used directly:

``` C++
    GuardedDownloadQueue downloadQueue = DownloadQueue::Guard();
    for (NzbInfo* nzbInfo : downloadQueue->GetQueue())
    {
        postJobCount += nzbInfo->GetPostInfo() ? 1 : 0;
    }
```

And use name **guard** if the guarded pointer is not used itself (but the lock is used to access objects within queue):

``` C++
    CString nzbName;
    {
        GuardedDownloadQueue guard = DownloadQueue::Guard();
        nzbName = m_fileInfo->GetNzbInfo()->GetName();
    }
```

### E-Mail address

Have you seen my [comment](https://github.com/nzbget/nzbget/pull/190#issuecomment-199828191) to another PR regarding your e-mail address?

hugbug:
> Have you seen my comment to another PR regarding your e-mail address?

I guess you did as the server volume commit is now properly linked to your account.

heksesang:
Updated the pull request to not include std::mutex changes, and updated the names of the guards.

PR #211, opened 2016-04-29
Prevent a crash on android > 5.x

Cpasjuste:
This prevent a crash on android running > 5.x. This also allow using untouched nzbget sources as a git submodule for android compilation with android-studio. Full sources will be up in a few days.


hugbug:
Thank you for contribution.

There is an issue with this change. All processes started by nzbget will not have system env. vars passed. I would rather prefer to investigate what is wrong with "g_EnvironmentVariables".

I don't have android 5 devices but I could try to debug in emulator if there is one for android 5 (don't know much about android development).

Cpasjuste:
Hi hubgub,

I fixed the android part by "mallocing" "argp". This pull request is no longer required. I should update the android repo in a few days.

Thanks,
David.

hugbug:
Great.

PR #212, opened 2016-04-29
Update nzbget.conf

LuisFlores3:


hugbug:
Thank you for contribution.

I'm sure there is many more texts to fix or rewrite. How about that?

LuisFlores3:
I will fix things as i stumble upon them. thanks for all you do. :0)

hugbug:
Great, I'll wait then. Tell me when you have it finished.
And I'd rather prefer to put all changes into one commit instead of doing one commit per word :wink:

war59312:
Few more:

https://github.com/war59312/nzbget/pull/1

LuisFlores3:
Okay, so here is a first pass, mainly just spelling, and hyphens and such. Two notes:  
- Line 499, i think you wanted 'omitted', not 'limited'. (I changed)
- Lines 1070-1105, I am not sure if 'outputmode' is an internal referance (good as is) or if it is supposed to be 'output mode' (conversational), same with 'frontend' (vs. 'front end') (I left as-is)

hugbug:
Great!
Should I merge or better wait for more corrections?

LuisFlores3:
I'm not sure when i will have more time to take a better look at grammar and the like. It's a bigger thing. so it might be a couple of weeks... up to you really.

hugbug:
OK, I'll merge it then (tomorrow).

PR #222, opened 2016-05-25
 @JackDandy  Fix incorrect variable. 

JackDandy:


hugbug:
Thanks but what's the issue with `true` and `false` and why I should use `!0` and `!1` instead? That's very unintuitive and I'm unhappy with the change.
Can you explain?

JackDandy:
Copied from commit text...

```
Fix the switch on Settings/Web-Interface/ActivityAnimation to allow play/pause button anim to be disabled.

Also,
Fix String types being compared to Boolean types assigned as default values.
Change to use short Boolean notation to prevent further issues arising from long notation misuse.
```

In all browsers on every device platform, !1 **evaluates** to absolute false and !0 **evaluates** to absolute true, look inside jquery code to observe a big player using the notation. By design, its use  prevents exactly the common _Type_ conversion mistake that I found. The additional value is that for inline code switches, only one character need be changed to switch the value, e.g. var debug = !0 or !1 versus the very cumbersome typing of fasle, i mean, falal, i mean false and true ;)

hugbug:
Using encrypted `!1` and `!0` is just insane. I understand that `!1` evaluates to `false` and `!0` to `true` but if the motivation is to save a few characters I'm not buying that. I'm totally ok with typing whole `false` and `true`.

> look inside jquery code to observe a big player using the notation

Where did you saw that in jQuery code? What I see is that they use `false` and `true` [everywhere](https://github.com/jquery/jquery/search?l=javascript&q=false&type=Code&utf8=✓).

> its use prevents exactly the common Type conversion mistake that I found. 

Can you please point me to the erroneous code?

> Fix String types being compared to Boolean types assigned as default values.

Do you mean code like this:

``` javascript
this.refreshAnimation = this.read('RefreshAnimation', this.refreshAnimation) == 'true';
```

Function `this.read` returns string. So the string is compared to string. No errors here.

JackDandy:
Ok here is one of _many_ examples, the following link is to the official Google hosted jquery library that serves many huge sites... `https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js` the usage of !0 and !1 is abundant in that file.  The main Google hosted page that presents the above and other libraries that they recommended is directly hot-linked into website pages is... `https://developers.google.com/speed/libraries/#jquery` and these files are linked vanilla in Wordpress installations and many other websites. I hope that helps clear up that what I was trying to illustrate isn't something I made up, indeed, I learned about it in a **paper** from the **ECMA** devs.

The motivation in the paper is certainly not to save a few chars and is certainly not a form of encryption, but nevermind, you are clearly not comfortable with the paradigm or in what I tried to show, no problem, I reverted except for the obvious incorrect variable name.

hugbug:
First of all let me thank you for taking your time to prepare the PR, discuss my questions and rework the PR. Really appreciate this 👍.

> https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js

That's a minified version of the library. It was produced by a special software (minifier) from the original code (which uses `true` and `false`) with a purpose to make the code smaller for hosting.

JackDandy:
Yup, it's minified and uses techniques presented from ECMA international, there are reasons they recommended using !1 and !0 paradigm and it certainly was **NOT** to save bytes - that was simply a lucky byproduct - but what do they know right? lol, gl. :)

hugbug:
> it certainly was NOT to save bytes 

It is indeed to save bytes - the only point of minification in the first place.

Take a look at the source code of [jQuery](https://github.com/jquery/jquery/search?l=javascript&q=false&type=Code&utf8=✓) and you will not see any `!0` or `!1` there because they use `true` and `false`.

> techniques presented from ECMA international

Any reference where ECMA (or anyone else) recommends `!0` and `!1` over `true` and `false`?

JackDandy:
Of course the point of minification is to save bytes, but that is NOT the point of the !0 and !1 paradigm as presented by ECMA. I helped to explain some reasons already, **one is to prevent obj Type misuse**, which you have.  You would have to Google, I read the paper long ago. But why are my words not enough? I'm not saying you must believe in something you cannot see, oh wait, you actually can not see it. lol :tease:

hugbug:
Your main argument was that jquery devs use this technique but they don't. Anyone else?

I've tried to google but what are the keywords? Google doesn't like a search for "!0".

What is "Type misuse" and where do I have it? If there is an issue with the code I do want to fix it.

> believe in something you cannot see

That's a good point, I indeed don't see anyone using or recommending this technique. You can't seriously count automatically generated minified code as "usage". Do you have a reference to any (preferably well known) project using "!1" and "!0" in their source base?

JackDandy:
Huh? your argument about jquery vs minified code is like trying to compare source with byte code, jquery source base must be more "readable" because it is a shared library used by banks, governments etc, where ng is an entirely different proposition, and btw, jquery have their own minifier much as yahoo, google and others do - all with different output.  No, to be clear, my point was simply that the use of !0 and !1 in JS is widespread, was recommended by ECMA, and is nothing to warrant your initial reaction, but now you suggest I go looking for others that use this? Okay, just because a majority do something a particular way doesn't mean everyone follows suit, that is hardly creative or engineering, I know there will be projects that use it in their source base - i am not the inventor nor the only one to have read the paper, but it is futile to go searching it out to prove what exactly!?, that there are few that know and use this advanced yet simple technique at the repo level? no, that's silly. I'll research it for a bounty though ;)

> You can't seriously count automatically generated minified code as "usage".   

Generated code is simply a pattern from human input repeated over, **where would you place the most opimised or best formed code conventions if not in a compiler or minifier? Surely, you would place it at source level if possible**, so yes, usage is usage, its all man made - not magic.

Back to the code at hand, I suggest you check the obj type of `this.activityAnimation` and how that object is subsequently misused. That is one of several cases in that file that if you can see what is actually wrong, I can explain why it would have been impossible to occur if !0 and !1 were used,  and the main point, which is to prevent bugs by code design so that thought isn't reqd. That is, if you even care to read more of my mutterings and musings :)

hugbug:
`!0` evaluates to `true` and is therefore an equivalent to `true`, it's not more boolean than `true`. The same about `!1`and `false`. There are also no such things as "absolute true" or "absolute false". `!0` isn't more robust than `true`, it just a different way of typing `true`.

Now, since `!0` and `true` are equal there is no way that using `!0` helps to avoid any kind of errors. And considering the cryptic nature of `!0`-notation comparing to clear `true` I would argue that using `!0` in code is more prone to errors.

> the use of !0 and !1 in JS is widespread, has been recommended by ECMA

I wouldn't ask about references if I could find anything myself. I did the search and failed to find **anyone** using or recommending this technique. That's a second reason why I'm skeptical.

You don't have to prove anything, we are just in a friendly discussion. We can stop it anytime.

> I suggest you check the obj type of this.activityAnimation 

I guess you mean how it is (automatically) converted to string and under certain conditions booleans are compared withs strings (which contain text "true" or "false"). It's intentional; that's why I use `==` here, instead of `===` which I use otherwise. `!0` and `!1` wouldn't change a thing, since they are the same as `true` and `false`.

JackDandy:
`typeof(true) != typeof("true")` and `false = (typeof(true) == typeof("true"))` the word "absolute" was used indicate a real Boolean over "true" / "false", nothing more.

Hoping for automatic type conversion of "true" to equal Boolean true is at best a very bad practice at worst will cause a situation where the condition (== "true") can never be anything but negative therefore is misuse.

> !0 and !1 wouldn't change a thing

Using strings this way is way bad and !0 and !1 guarantees this can never occur - simple, elegant, no bugs by design, so yes, it fundamentally changes everything because it enforces bug free code comparison with Booleans, there is little chance you would wrap !1 in quotes, but look how easy it was to do with the word "true", why write using risky constructs like that!? 
Bottom line, !0 >== true and !1 >== false ;)

hugbug:
Thanks for the explanation. Now I see the problem with "this.activityAnimation". The default value must be passed as string, not as boolean, to avoid comparison of boolean to string if the storage doesn't have a value. Alternatively the read-function must convert the default value to string before returning.

Still `!0` wouldn't help to avoid the problem as I would compare to "true" anyway since that's what function `read` returns (string "true" or "false").

JackDandy:
And still you insist on refuting the legitimacy of !0 and !1 lol, gl.

hugbug:
The local storage saves and reads only strings. Even if I would write `!0` it would return `"true"` on reading. So I need to compare with string `"true"` anyway. You see, here `!0` doesn't help to avoid the bug :wink: .

JackDandy:
Okay, one last effort, it's **all** about training your mind to code in a certain way so that certain issues cannot occur by design - this is key - you need to appreciate this first otherwise you've already failed.

Now, because `this.read()` returns a string "true" or "false" from, in this case, localStorage, you are unfortunately immediately susceptible to writing buggy code as your mind naturally thought to compare the return string equals "true" and you forgot about the var declared boolean type. **If instead, you always use !0 and !1, then you would not have to think, worry or as in this case _forget_ about the variable declaration being of type Bool because it is abundantly clear by using `== !0` instead of `== 'true'` or `== true` as you would never think to quote `== !0`**. Instead you would naturally type cast the `this.read()` function to return Bool - **therefore, by design you would NOT have created the bug in the first place** - and all without even thinking about it.
And to the other issue of returning the correct type _from input_ in `this.read()`...

```
else if ('true' === v || 'false' === v)
    return 'true' === v;  /* returns Bool cast from string */
else
    return v;  /* perfectly fine that this returns type string as usual */
```

And yes, absolutely, this usage reasoning was given by ECMA in the paper to a similar example, now you can see why this isn't something easily found in search engines. A simple but huge effort saver.

hugbug:
Function `this.read` returns a value from local storage. It must always return a string because it doesn't know the type of value. It's a responsibility of the caller to interpret the string value. If `this.read` would try to return differen types as you suggest, it may happen that it returns a Boolean for setting `windowTitle` for example; that will likely cause troubles at places where `windowTitle` is used.

A mistake in `this.read` is that it may blindly return default value if no value is in local storage. Since it is supposed to return strings in all cases it's better to convert all default values to strings before returning.

Now to work with booleans we need to convert them to/from strings. Function `this.write` accepts all types and convert them to strings automatically. We can use automatic conversion from boolean to string or we could do our own. When using `this.read` with non-string types we have to exlicitly convert from string to other types. For setting `refreshInterval` we use `parseFloat`. For booleans we use string comparision with `"true"`. It's important to note that we compare string with string here. A mistake here is that under certain conditions boolean is compared with string. That has to be fixed.

The `!0` technique wouldn't help here as we need to check string values `"true"` and `"false"`. OK, you could say **never use strings "true" and "false"**. That would force us to use some conversion when saving booleans as strings (for example we could use "1" and "0") and later to interpret them when reading. Doing so would require to explicitly convert default value to string **before** passing to `this.read` and would eliminate the mistake before it could happen.

Agreed but **never use strings "true" and "false"** and **always use !0 and !1 instead of true and false** are two very different practices. I would say hardly related to each other.

JackDandy:
> It's a responsibility of the caller to interpret the string value

Nope, data should be validated (and Type cast) at the point of being read from the input source - this is not only standard security practices, but also creates a single point of code maintenance and prevents other issues too.

>  It must always return a string because it doesn't know the type of value.

and

> A mistake in this.read is that it may blindly return default value if no value is in local storage. Since it is supposed to return strings in all cases it's better to convert all default values to strings before returning.

There is one adjustment to validate the input from localStorage...
`else if (typeof(!0) === typeof(def) && ('true' === v || 'false' === v))`
This holds the integrity by matching the obj type of **def**ault in cases where the callee may actually require a string that equals "true" or "false" to begin with ;)

With my suggestion, when the callee is requesting a bool value (by the fact that `def` is bool), then `this.read()` will return a boolean (see the above tweak). For anything else, a string or "undefined" can be returned, there is nothing to say that a function must return only one variable type. Again, because the function param **def**ault exists for `this.read()`, then this conveniently provides the return type to fulfill.

The mistake is that the data type isn't validated at the point of coming in from localStorage against what is expected to be fulfilled. All that is required is; identify and return the correct type to the callee. Nothing more complicated.

> That would force us to use some conversion when saving booleans as strings (for example we could use "1" and "0")

and 

> later to interpret them when reading.

and

> Doing so would require to explicitly convert default value to string before passing to this.read

Compared to only using !0 and !1 at the callee??  Hmm okay, strings '0' and '1' also bring Type issues.  All that faffing around in order to avoid the blatantly simple use of !0 and !1 which **explicitly evaluates** as a singular boolean type, and compares (in if/while etc) the same on every device, and is not vague like how '0' can be numeric and yet can also be false etc, !0/!1 locks out any issues without conversion, without needing to "_interpret them when reading_", and without even thinking about the usage.  A change to the saved string values is futile and displays nicer in browsers as "true" and "false" (e.g. the value appears as tooltips in mine).

> The !0 technique wouldn't help here as we need to check string values

and 

> Agreed but never use strings "true" and "false" and always use !0 and !1 instead of true and false are two very different practices. I would say hardly related to each other.

Anyway, you seem to completely fail to appreciate the point of the design, ah well.

hugbug:
Checking the type of default value and treating it as requested return type is indeed another possibility. But this again isn't related to `!0`.

JackDandy:
It is all related to a design paradigm, be in denial, you have been shown the light :p

hugbug:
> With my suggestion, when the callee is requesting a bool value (by the fact that def is bool), then this.read() will return a boolean (see the above tweak). For anything else, a string or "undefined" can be returned, there is nothing to say that a function must return only one variable type. Again, because the function param default exists for this.read(), then this neatly set up what return type to fulfill.
> 
> The mistake is that the data type isn't validated at the point of coming in from localStorage against what is expected to be fulfilled. All that is required is to identify and return the correct type to the callee. Nothing more complicated.

It's a design question wether `read` or the caller should make type conversion. Both alternatives are valid. Yet this has nothing to do with **`!0` vs `true` discussion**.

> simple use of !0 and !1 which explicitly evaluates as a singular boolean type
> 
> !0/!1 locks out any issues without conversion

`true` and `false` are not any worse in this respect. The problem in your arguments is that you propose a design change to function `read`, a design change which is not related to using of `!0`. Then you use that proposal to defend the `!0`-technique although the core of the proposed change is to check the type of the default argument and to to use `!0` instead of `true`.

I'm not arguing that `read` can be implemented differently but you still didn't show how `!0` is better that `true`.

You also defend the technique by saying it's **widely used and recommended by ECMA** yet failed to provide a single prove for that. First you appealed to jQuery, then when I've shown that they don't use it you argued that jQuery needs readability in the source base. As if readability were a bad thing which I don't need.

---

Anyway, thanks for the discussion. You pointed me to a bug which has to be fixed and that's a good outcome 😉.

P. S. it's harder to follow the discussion if the posts are edited.

JackDandy:
About the edits, perhaps i should use a texted and then copy paste, but this isn't twitter.

> you still didn't show how !0 is better that true.

Yes I did, read back - its mainly about **preventing misuse by design. The misuse was proven in how the code was written by using `true` which later got compared to `'true'`, this quotation leaked into or from not using the correct type in `read()` and would not have occurred if using `!0` in the first place - that is one of few reason that I gave as to why using `!0` is better than `true` in JS source**. The proof of bug involved the function `read()`, so it all served neatly as a real life proof example rather than something anecdotal.

> failed to provide a single prove for that

I showed exactly where it is used, and you disagree saying it is invalid because a human made program exchanged "true" into !0, which does not invalidate my proof. I also said that given a paradigm, one can use it where ever.

> As if readability were a bad thing which I don't need.

Readablity is indeed a good thing, however, my point was that its sometimes required, sometimes not. The unfamiliar does not prove unreadability and there is nothing unreadable with `var debug = !0` :wink:.

hugbug:
> Yes I did, read back - its all about preventing misuse by design. The misuse was proven in how the code was written by using true which later got compared to "true" and this quotation would not have occurred if using !0 in the first place. 

I don't think so. It was compared with `"true"` because local storage saves and reads strings. Even if you pass `!0` to `localStorage.setItem` you end up reading `"true"` from `localStorage.getItem`. And in your changed code you indeed compare with string `"true"`, just like I did, but in a different code place (not directly in `read`). That's why I'm not considering this as a proof of how `!0` helps avoiding bugs.

> Readablity is indeed a good thing

I'm glad we at least can agree on that :beer:.

JackDandy:
> Even if you pass !0 to localStorage.setItem

As I said, "A change to the saved string values is futile"

> It was compared with "true" because local storage saves and reads strings.

Exactly, i said this already.

> You indeed compare with string "true".

Absolutely, and...
1) I am doing it only **once** on the input validation in `this.read()`. The rest of the code base would be on lock down by virtue of not using it, whereas prior, you had the same error multiple places in code all of which would have been avoided with !0 and a focus in only one place for the careful use of "true".
2) that _singular_ check of localStorage data can be further cross browser strengthened...

```
            return def.toString();
        }
        else if (typeof(!0) === typeof(def)) {
            var v = v.toLowerCase(), bItemTrue = ('true' === v);
            if (bItemTrue || ('false' === v))
                return bItemTrue;
            this.ng_debug && console.log('browser storage failed to return known boolean type');
            return def;
        }
        else
```

[edit] theory code fixed (1.Jun)

 :wink:

hugbug:
Instead of `typeof(!0)` we can write `typeof(true)` making the code more readable, right?

I personally prefer to keep `read` simple. You added the conversion logic for booleans but forget to add a conversion for integers (used by two other settings). Anyway the decision where to convert the string (in `read` or outside) is not related to the usage of `!0`.

JackDandy:
It's not a conversion or coercion - it's an underused fundamental data type that is not 0 or not 1 which evaluates exactly equal to the representation of true or false on any environment.  You can add similar type checks to floats, ints or prototyped objects, I left that for you to handle after suggesting where to place such code -- as I said, data validation principles (listed at owasp) recommend to put integrity checks at the boundary of reading input - this model serves to isolate security code so the dev can proceed knowing that a variable value is already "safe" when using it. Also it is a single point of change if an new issue is found.

Good luck.

hugbug:
> it's an underused fundamental base data type that is not 0 or not 1 

`!0` is not a data type, it's an expression which evaluates to boolean and after evaluation is equal to `true`. Everywhere where you suggest to write `!0` we can write `true` and achieve better code readability, without loosing anything at all.

Your arguments about where to put string conversion or validation are absolutely unrelated to the topic **`!0` vs `true`**. 

I'm not arguing about where to convert `"true"` to boolean. I'm asking about why use `!0` instead of `true`? You haven't provided a valid argument for that. The question about who else uses or recommends the technique remained unanswered as well.

JackDandy:
---

Hmm, okay, we should draw a line here and agree to disagree, unless you want to disagree to disagree which then becomes directly proportional to some level of discombobulation :wink:

hugbug:
Yeah 👍

Let me assure you that I'm thankful to you for the whole discussion and in particular for two bugs you've discovered. And I'm also learned that comparison  `boolvar == "true"` is a big mistake.

You see, I'm convincible 😉.

PR #255, opened 2016-08-04
Do not compile ParCheckerTest and ParRenamerTest if parcheck is disabled

gktrk:


hugbug:
Oh, missed that. 
Thanks a lot for the fix and generally for contribution.

PR #256, opened 2016-08-04
GCC 4.8 (C++11) support

caronc:
With respect to a [blog](http://nuxref.com/2016/08/03/nzbget-v17-centos-7/) I wrote on NZBGet v17, the only complaint I had with it was the fact everything was re-factored to use the C++14 standards to which most (Stable) Linux distributions don't support. I was bummed out when i found out i couldn't compile it anymore with CentOS 7 specifically.

I wrote a patch that made it compatible with my OS (by reverting a bunch of the code to follow C++11 standards instead) [and shared it here](http://repo.lead2gold.org/pub/patches/nzbget/nzbget-use.c++11.patch).  Although I may have started the initiative of this change, @hugbug gets credit for _greatly_ simplifying the patch I had started and sharing it with me for this commit.

This pull request provides the ability to compile NZBGet on older Linux Distributions for those (like me) who want to build it themselves and have it link to the shared libraries maintained by our OS.

### DIFF
--- a/daemon/main/Scheduler.cpp
+++ b/daemon/main/Scheduler.cpp
@@ -40,7 +40,7 @@ void Scheduler::FirstCheck()
 		Guard guard(m_taskListMutex);
 
 		std::sort(m_taskList.begin(), m_taskList.end(),
-			[](std::unique_ptr<Task>& task1, std::unique_ptr<Task>& task2)
+			[](const std::unique_ptr<Task>& task1, const std::unique_ptr<Task>& task2)
 		{
 			return (task1->m_hours < task2->m_hours) ||
 				((task1->m_hours == task2->m_hours) && (task1->m_minutes < task2->m_minutes));
--- a/daemon/main/nzbget.h
+++ b/daemon/main/nzbget.h
@@ -340,4 +340,19 @@ typedef unsigned char uchar;
 #define SCANF_SYNTAX(strindex)
 #endif
 
+// providing "std::make_unique" for GCC 4.8.x (only 4.8.x)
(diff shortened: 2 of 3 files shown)


Prinz23:
At least on CentOS 7 gcc 4.9.2 should work according to this: https://gist.github.com/craigminihan/b23c06afd9073ec32e0c

caronc:
Except, what you're implying is that we don't adopt and use the mainstream packaged versions which are patched and maintained regularly by Red Hat (and then by the CentOS community). Compiling and manually installing a new version of the glib core ourselves forfeits all the stability we chose the OS for in the first place. @Prinz23, I guess what I'm saying is that a link to some ghist of someone who hacked their system and destroyed any integrity they had isn't a good argument. The proposal isn't a re-write of NZBGet, it's just more compiler support for non-bleeding edge systems.

hugbug:
Thank you.

@Prinz23: considering the very small amount of changes in the patch, it't a good thing to have nzbget compatible with gcc 4.8 out of the box. Especially for users of CentOS, who has no option to upgrade their OS at the moment, unlike, for example, users of Ubuntu and Debian.

hugbug:
## Documentation

_to merge into wiki after v17.1 release_

### Usage

To compile with gcc 4.8 use the following configure command:

``` shell
CXXFLAGS="-std=c++11 -O2 -s" ./configure --disable-cpp-check
```

### Important notes:
- compiling is possible only without debug info. When trying to compile with `-g` the compiler fails with internal error. The [original patch](http://nuxref.com/2016/08/03/nzbget-v17-centos-7/) works in debug mode too but brings additional changes to source code.
- support for gcc 4.8 is considered temporary and may be dropped in future NZBGet versions.

Issue #49, opened 2015-07-10
Password fields should be type="password"


mven:
It's currently:

``` html
<input type="text" id="S_Server1_Password" value="" class="editsmall">
```

It should be:

``` html
<input type="password" id="S_Server1_Password" value="" class="editsmall">
```

hugbug:
In my opinion it's better to have password fields plain because that makes it easier to control the settings.
On the other side I understand your concerns.

**Suggestion:** show passwords masked by default but provide a way to show them as plain text. Two alternatives:
1. button "Show" near each password field;
2. option "Show passwords" in menu "View" on settings. That menu currently have only one option "Compact mode".

I'm adding label "help wanted" to this issue. Contributions are welcome.

marcelbrueckner:
+1 for type="password"

IvanJosipovic:
+1 for type=password and a checkbox for "Show Passwords"

RollingStar:
Is this just for managing Usenet servers? Does it apply at all to NZB download passwords?

hugbug:
It's not for nzb-passwords.

cxhamilton:
What is the problem that you are trying to solve? If the problem is someone looking over your shoulder, the masking or hiding the password with an option to reveal would be suitable. If the problem is a proxy or "man in the middle" inspecting your unencrypted traffic, the only secure option would be to remove the feature that shows a stored password. The password field would be one way, to set or update a password.

war59312:
Well +1 as it helps with password managers. ;)

Should also help with screen readers like jaws to recognize it's a password field.

Speaking of... Anyone ever test NzbGet with screen readers. Just because your blind, should not mean you can't use NzbGet.

guaguasi:
Just created PR #260 for this.

PR #260, opened 2016-08-12
Fix for #49 (Password fields should be type="password")

guaguasi:
Addresses #49. Done in 2 commits in case you want to support archive passwords as well.


hugbug:
Very nice. Thank you.

I'm currently preparing v17.1 as a bugfix release for v17.0. Once it's done I'll merge your PR. That may take a couple of weeks (including intermediate 17.1-testing release) but don't worry; consider the PR accepted.

Issue #29, opened 2015-06-22
Hours+minutes in custom pause


hugbug:
The temporary pause dialog currently accepts only minutes:
![screen shot 2015-06-22 at 22 52 48](https://cloud.githubusercontent.com/assets/3368402/8292721/70d25392-1931-11e5-8b31-9168dad81067.png)

Being able to enter hours+minutes would be helpful. Two cases:
- pause for **hh:mm**;
- pause until **hh:mm**.

Forum topic [Add Hours to Custom Pause Time](http://nzbget.net/forum/viewtopic.php?f=10&t=1666).

teonar:
Do you want  separate entry boxes ? 
Or do you want the <b>for</b> to be a dropdown box with choices (<b>for/until</b>)

hugbug:
I didn't though much about this yet but how about this:
- Only one edit field:
  - **30** - pause for 30 minutes;
  - **1:30** - pause for 1 hour 30 minutes;
  - **1h** - pause for one hour;
  - **=16:30** - pause until 16:30;
  - **=16h** - pause until 16:00;
  - **=16** - pause until 16:00. it doesn't make sense to pause until xx minutes, therefore the suffix "h" is optional in that case;
-  For the "pause until" we probably need to support both 24 hours and 12 hours am/pm format for user from different countries.
- The suffix "minutes" after the edit field must be removed.
- Below the edit field add a description text with few examples.
- When the user starts typing replace the description text with info about how the input was interpreted. For example, the user typed "=16", the description shows "Pause until 16:00". That's optional, if you think it's overkill :blush:.

guaguasi:
I've done this. I'm cleaning up the code and will have a PR tomorrow.

PR #278, opened 2016-09-10
Fix for Issue #29

guaguasi:
Took a little longer than I planned. Refactored twice. Hopefully worth it lol.

Adds more robust pause control as requested in issue #29

### DIFF
--- a/webui/index.html
+++ b/webui/index.html
@@ -897,7 +897,7 @@
 						<input class="input-mini" id="PauseForInput" size="16" type="text">
 					</div>
 				</div>
-				<p id="PauseForPreview" class="text-center" style="display:none">
+				<p id="PauseForPreview" class="text-center invisible">
 					Will remain paused until <strong></strong>
 				</p>
 				<hr />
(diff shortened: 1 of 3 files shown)


hugbug:
Great job you did!

I have one suggestion - the table with examples is useful but scares a bit (too much info). I would prefer a one sentence help (something like `For example "30m", "2h" or "=18:00"`) and button "Quick help" which shows the full help screen in a separate dialog, like in the "Add files dialog".

guaguasi:
The examples are really the only formats accepted. I'll do as you say and add the table in a "Quick help" style section and that sentence where the table is currently.

I'll get to it when I have some time, most likely this weekend. If I can manage, I'll do it tonight or tomorrow night.

hugbug:
A small thing: when the preview text appears or hides the dialog height changes. It'd be better to keep the height constant. May be some explanaition text in the preview box or just keep it empty but still visible to prevent height changes.

I've also noticed that opening the dialog for second time shows a preview text from previous usage.

guaguasi:
Ended up just doing it. Let me know if there are anything else.

Issue #282, opened 2016-09-16
Checkbox at the top left of the History table


JackDandy:
Ref: **NZBGet 18.0-testing**

Hello,

The checkbox at the top left of the History table, the one that selects all items on the page - ah, no it actually selects all items on all the pages that you cannot even see. 

This is really bad, I just wiped and lost an entire 640 record history for an item I was investigating for an issue, all that history gone when I expected to just be deleting records on the current page. Please change this behavior to be defacto normal - i.e. to select what is currently visible on screen.

Thanks

guaguasi:
Just submitted a PR for this.  I didn't realized this myself, so thanks for bringing it up!

hugbug:
I'm sorry for the loss of your history.

However I don't think your suggestion is an improvement. I don't see why anyone would **ever** need to select all items on current page; there is no real use case for that. On the other side selecting of all items is useful. Usually you either select certain items by clicking on their checkboxes or you select all items.

The second is in particular useful in combination with search/filter functionality. For example, let's say you need all items for a certain show. You type show name in the search box and the table shows all episodes. Would you want to select all episodes on the first page shown? No. You would either select all of them or a few particular ones.

This means you suggest to remove a very useful function and replace it with a useless one.

If you need to select many items and they happen to be all on one page you can use the selection for ranges: click on first item, then shift+click on the last item.

guaguasi:
You do have a point, but I do feel that there might be use cases such as cleaning up records page by page and saving yourself a scroll and shift click. I changed the code to keep the original functionality as the default and the modifier selects only visible items for the edge cases.

hugbug:
As an alternative solution for the original problem of accidental deleting of the items: what if the pages on the pager would indicate that they have selected items? For example with the same yellow background used in the table for selected items. Would you notice that?

JackDandy:
I understand your feelings but you're missing the point - **nowhere** in ui designs does a checkbox select items that are are not visible. Therefore, complete surprise comes when NG selects all items that no app or web based selector selects. With no undo for the destructive delete operation this is a double kicker.

An example, go look at any email web app, the first app type many learn to use, it doesn't select all the items ever with that top checkbox, it selects visible records - that is the standard defacto behavior. What NG is doing is in opposition to what the user expects from learning's over years across many apps.

[...]

PR #283, opened 2016-09-16
Select-all checkbox

guaguasi:
Default functionality is now selecting only visible. If clicked with a modifier, Control or Windows/Command, it uses it's original functionality.

Addresses #282


JackDandy (commented):
perhaps add to the input element something along the lines... [code]title="use modifier (ctrl on Win, cmd on osX) to apply action to all items not just the ones on this page"[/code] so that the user is made aware of the modifier

guaguasi:
I'm closing this for now because of the discussion and that there are bugs I haven't had a chance to address.

Issue #274, opened 2016-09-09
Show password-badge for nzbs with passwords


hugbug:
For nzbs having password set on tab "postprocess" it would be nice to have badge "passworded", "protected", "encrypted" or similar (not sure which one) displayed in the queue and history lists.

A mouseover-hint could show the password itself.

guaguasi:
I just did this.

How would you want mouseover-hint done? I thought about using the bootstrap tooltip, but I'd have to write a click handler and manually launch the tooltip since Bootstrap 2.0.3 doesn't have any other trigger options. The placement would also need to be changed depending on the layout. Not exactly the most  elegant, but easily doable. I could also just append `: <password>` to the label on click (and/or hover). Simplest solution would be to just add a title attribute but that wouldn't be visible on mobile.

hugbug:
I meant the title-attribute. Doesn't work on mobile, yes, but that's just a little bonus feature for desktop.

PR #285, opened 2016-09-20
Show password-badge for nzbs with passwords (#274)

guaguasi:
Label in blue denoting "encrypted" file with password displayed via title attribute on mouse over.

### DIFF
--- a/webui/downloads.js
+++ b/webui/downloads.js
@@ -771,7 +771,8 @@ var DownloadsUI = (new function($)
 		}
 	}
 
-	this.buildEncryptedLabel = function(parameters) {
+	this.buildEncryptedLabel = function(parameters)
+	{
 		var isEncrypted = false;
 
 		for (var i=0; i < parameters.length; i++) {
@@ -783,7 +784,7 @@ var DownloadsUI = (new function($)
 			}
 		}
 		return isEncrypted !== false ?
-			' <span class="label label-info" title="'+ isEncrypted +'">ENCRYPTED</span>' : '';
+			' <span class="label label-info" title="'+ isEncrypted +'">encrypted</span>' : '';
 	}
 
 	function formatDupeText(dupeKey, dupeScore, dupeMode)
--- a/webui/downloads.js
+++ b/webui/downloads.js
@@ -784,7 +784,7 @@ var DownloadsUI = (new function($)
 			}
 		}
 		return isEncrypted !== false ?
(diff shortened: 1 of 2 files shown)


webui/downloads.js
```diff
@@ -769,7 +770,22 @@ var DownloadsUI = (new function($)
...
+		for (var i=0; i < parameters.length; i++) {
+			if (parameters[i]['Name'] === '*Unpack:Password' &&
+				parameters[i]['Value'])
+			{
+				isEncrypted = parameters[i]['Value'];
+				break;
+			}
+		}
+		return isEncrypted !== false ?
+			' <span class="label label-info" title="'+ isEncrypted +'">ENCRYPTED</span>' : '';
```
hugbug: Use `Util.textToAttr(isEncrypted)`.
Also please use lower case for "ENCRYPTED" like for all other badges.

webui/downloads.js
```diff
@@ -769,7 +770,22 @@ var DownloadsUI = (new function($)
...
-	
+
+	this.buildEncryptedLabel = function(parameters) {
+		var isEncrypted = false;
+
+		for (var i=0; i < parameters.length; i++) {
+			if (parameters[i]['Name'] === '*Unpack:Password' &&
+				parameters[i]['Value'])
+			{
+				isEncrypted = parameters[i]['Value'];
```
hugbug: I don't understand why you assign string variable to boolean?

webui/downloads.js
```diff
@@ -769,7 +770,22 @@ var DownloadsUI = (new function($)
...
+				break;
+			}
+		}
+		return isEncrypted !== false ?
```
hugbug: Here isEncrypted can be either boolean (false) or a string. Is it ok to compare string with `false`?

webui/downloads.js
```diff
@@ -769,7 +770,22 @@ var DownloadsUI = (new function($)
 			return ' <span class="label label-priority label-info">priority: ' + priority + '</span>';
 		}
 	}
-	
+
+	this.buildEncryptedLabel = function(parameters) {
```
hugbug: please don't use dangling braces `{` for format consistency.

hugbug:
Nice done BTW 👍
I hope the comments don't sound too harsh.

guaguasi:
Not at all. lol.

I had originally started not having a label title but changed it midway and just never renamed the variable. You're right about the `false` comparison. `encryptedPassword` is only ever going to be undefined, a blank string, or an actual password. My original intent was to account for other falsy values but then I realized I'm dumb.

PR #311, opened 2016-11-21
Update README.md

rotatordisk92:
Added gitter.im for developer discussions


hugbug:
As I suspected the chat isn't that useful and GitHub issue tracker and NZBGet forum remain the recommended platforms for communication.

Issue #320, opened 2016-12-20
nzbget nntp server: slowness in NNTP session due to missing newline after AUTHINFO response


sanderjo:
**Short**: slowness in NNTPServer-client-session due to missing newline after AUTHINFO -> 281 response

**Setup & effect:** Pure nzbget setup: client & server (nzbget 18.0-testing-r1841). Client has username/password set for localhost-newsserver, and that causes delays of 9 seconds in Test Connection. And I believe even longer timeouts leading to blocking trying to download.

**Demo**: https://www.youtube.com/watch?v=H1fKSYPjWJQ

**Workaround**: no username/password filled out on client side

**Real solution:** add \r\n like below

NntpServer.cpp (see https://github.com/nzbget/nzbget/blob/a98bbd7d0d8beeda3d41ff02e5b1de5ac9df0674/daemon/nserv/NntpServer.cpp )

```
		else if (!strncasecmp(line, "AUTHINFO ", 9))
		{
			m_connection->WriteLine("281 Authentication accepted");
		}
```

I think that line should read

```
			m_connection->WriteLine("281 Authentication accepted\r\n");
```

**Manual session shows the problem:**
```
$ telnet localhost 1190
Trying ::1...
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
200 Welcome (NServ)
BODY blabla
430 No Such Article Found (invalid message id format)
ARTICLE blabla
430 No Such Article Found (invalid message id format)
AUTHINFO user blabla
281 Authentication acceptedJKJKJKJKJKJKJKJKJ<ENTER>   <--- no Newline, so random typing with enter
500 Unknown command
````
Pen-ultimate line: no newline sent by nzbget ... :-(

**Bonus**

Same problem for:
```
m_connection->WriteLine("205 Connection closing");
```

sanderjo:
Pull Request sent...

Code tested on Ubuntu 16.10, and works fast now. :-)

hugbug:
Thank you for reporting this and for the fix of course. :+1:

PR #321, opened 2016-12-20
Newlines added to 281 and 205 server answers

sanderjo:
To solve delays described in https://github.com/nzbget/nzbget/issues/320


hugbug:
Thank you!

Issue #24, opened 2015-06-21
Double click handler on Windows


hugbug:
When clicking on nzb-files in Windows Explorer the files must be enqueued in NZBGet.

Current solution requires users to manually add a registry entry (with command **nzbget.exe -A %1**) and works only if NZBGet is currently running.

brosoft-dev:
The "NZBGet must be running" clause can be solved by hijacking the RemoteClient::RequestServerDownload() which (if I'm not mistaken) is where the "-A <file>" command line args end up... If InitConnection() fails to connect to the server, then we have the option to simply copy the NZB file to $(NzbDir) where it'll be picked up by Scanner:: once the daemon starts.

In the meantime (and having never seen NSI script before) - this is my attempt to give users double-click functionality for local installations... May need some work; but if you insert this somewhere in the Main section...

``` nsi
; Ask if we should add shell shortcuts
MessageBox MB_YESNO "Add NZB file actions to shell?" \
  IDNO noshell

; Are nzb files already defined? Hijack the regkey...
ReadRegStr $R3 HKCR ".nzb" ""
${If} $R3 == ""
  ; else, define our own
  StrCpy $R3 "nzbfile"
${EndIf}

; Set the file description type
WriteRegStr HKCU "$R3" "" "NZBGet File"

;; Add an action

; (only if other commands are present) this forces our command to be the default double-click action
;WriteRegStr HKCU "$R3\shell" "" "queue"
; Gives the file a default icon
WriteRegStr HKCU "$R3\DefaultIcon" "" "$INSTDIR\nzbget.exe,1"
; This defines the text you see on right-click
WriteRegStr HKCU "$R3\shell\queue" "" "Queue in NZBGet"
; This is the actual command the system runs when actioned...
WriteRegStr HKCU "$R3\shell\queue\command" "" "$\"$INSTDIR\nzbget.exe$\" -A $\"%1$\""

; Add a "Queue first in NZBGet" command (add to top of queue)
;WriteRegStr HKCU "$R3\shell\queue1st" "" "Queue first in NZBGet"
;
;WriteRegStr HKCU "$R3\shell\queue1st\command" "" "$\"$INSTDIR\nzbget.exe$\" -A T $\"%1$\""

noshell:
```

hugbug:
Thank you for your contribution.
1. I would happily merge it if you could make a proper pull request out of it. Provided it is fully functional and you have tested it of course :wink:.
[...]

PR #365, opened 2017-04-22
#24: Add option for file association

wtf911:
This will add a checkbox so after installation finishes there will be an option to associate .nzb files with NZBGet. It will also remove the association if NZBGet is uninstalled. This solves issue number #24.

### DIFF
--- a/windows/setup/nzbget-setup.nsi
+++ b/windows/setup/nzbget-setup.nsi
@@ -77,6 +77,8 @@ RequestExecutionLevel admin
 !insertmacro MUI_PAGE_LICENSE "..\NZBGet\COPYING"
 !insertmacro MUI_PAGE_DIRECTORY
 !insertmacro MUI_PAGE_INSTFILES
+!define MUI_PAGE_CUSTOMFUNCTION_SHOW MyFinishShow
+!define MUI_PAGE_CUSTOMFUNCTION_LEAVE MyFinishLeave
 !insertmacro MUI_PAGE_FINISH
 
 !insertmacro MUI_UNPAGE_WELCOME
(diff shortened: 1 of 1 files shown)


hugbug:
Thank you.

Uninstaller deletes ".nzb" key unconditionally. Wouldn't it be better to delete it only if it is set to "NZBGet.NZBFile", so that if user has changed association to another program the uninstaller wouldn't delete it?

wtf911:
Yes that would be better. :) I also still want to figure out how to have it open NZBGet if it isn't already running.

hugbug:
> I also still want to figure out how to have it open NZBGet if it isn't already running.

Instead of starting with `nzbget.exe -A %1` use another (new) switch to let nzbget uniquely identify that it was started with double click, for example `nzbget.exe -add %1`.

In [`WinConsole::InitAppMode`](https://github.com/nzbget/nzbget/blob/develop/daemon/windows/WinConsole.cpp#L103) check that switch and check if nzbget is already running (see  [`WinConsole::CheckRunning`](https://github.com/nzbget/nzbget/blob/develop/daemon/windows/WinConsole.cpp#L681)).

- if nzbget isn't running - modify the command line to start as server in application mode (with tray icon), `InitAppMode` has an example of that. Then enqueue nzb-file with [`Scanner::AddExternalFile`](https://github.com/nzbget/nzbget/blob/develop/daemon/queue/Scanner.cpp#L482). You may need to wait until nzbget is fully initialized, something like:
    ```C++
	while (!DownloadQueue::IsLoaded())
	{
		usleep(20 * 1000);
	}
    ```
   The waiting and enqueuing of nzb must be done in `WinConsole::Run` which runs in a separate thread. Doing all that in `InitAppMode` would block the program start.
  Note: the modification of command line is necessary so that on reloading via webui it wouldn't try to enqueue file again (it fully reinitializes and parses command line again);

- if nzbget is already running - send the nzb-file to running instance using [`RemoteClient`](https://github.com/nzbget/nzbget/blob/develop/daemon/main/nzbget.cpp#L759). That's how `nzbget.exe -A` would do that. Then terminate.

hugbug:
> if nzbget isn't running - modify the command line to start as server in application mode 

Or maybe instead launch another (server) instance of nzbget and wait until it initializes, then proceed as in "if nzbget is already running".

wtf911:
This goes over my head a bit haha. Learning to use NSIS and enough python to make the sorting script was one thing. Now I'm going to learn to use C++? :D Hmm. I can try but if I'm going to do it I might need more of a nudge in the right direction at some point.

hugbug:
>Now I'm going to learn to use C++?

I thought you know it ;)

I'm going to merge this PR in current state as it's already useful.

wtf911:
Nope, I don't know C++. Or any programming language proficiently haha. And sadly it looks like I won't be able to figure out how to add a new switch and get this working. I think I need your help. Thank you for letting me be a part of this. :)

PR #370, opened 2017-04-29
Document unlimited download rate setting

RollingStar:
I wasn't sure if the scheduler could set an unlimited download rate like you can in the downloadrate section on the top-left of the Web UI (here) 
image

A change like this makes things more obvious. I believe nzbget.conf governs the text in the WebUI: Scheduler section.


hugbug:
I've added a note about unlimited - see  https://github.com/nzbget/nzbget/commit/197baf066adf509eb3cfc960cec97703d179a57f. It's more like to other descriptions than the formatting you've used.
Thanks.

RollingStar:
Thank you, I wasn't sure how to write it.

PR #386, opened 2017-06-01
Don't write beyond buffer when reading a signature

sinic:


hugbug:
Thanks.
I wonder how did you catch this, did it crash for you during update? Have I uploaded bad signature file? What is your OS (more specifically which one signature was longer than expected)?

sinic:
No, I was just reading the code.

hugbug:
Cool.
I'll need to test the update function before merging; it will take some time.
Thanks again.

sinic:
Any plans for a release? I have a PoC that gets an attacker from MitM to RCE, so this is rather serious.

hugbug:
The fix will be released with v19, in the next couple of weeks.

When you talk about attack you don't mean through nzbget? Such attack would work only during update procedure via web-interface, which is a rare thing. And if you concerned you can update manually.

sinic:
Yes, I was only talking about the web-interface.

PR #400, opened 2017-06-24
Connection now supports UNIX sockets if host starts with a slash

schnusch:
- for people using nzbget behind a reverse HTTP proxy, it does not occupy a port on the loopback interface and provides better security, because access from other users on the same machine is no longer possible over a port on the loopback interface, but is now controlled by file permissions
- currently breaks CLI client but there are probably a couple more people out there who also basically never use it

### DIFF
--- a/daemon/connect/Connection.cpp
+++ b/daemon/connect/Connection.cpp
@@ -19,6 +19,9 @@
  */
 
 
+#ifndef WIN32
+	#include <sys/un.h>
+#endif
 #include "nzbget.h"
 #include "Connection.h"
 #include "Log.h"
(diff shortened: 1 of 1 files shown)


hugbug:
Thanks for the contribution.

I think that's pretty useless ;) but nonetheless I can integrate this if you make it less intrusive. My goal is to minimise the amount of code needed for the feature; the code I need to support in years to come.

I suggest the following changes:
- instead of new option **ControlSocket** (mis)use existing option **ControlIP**. For normal sockets it contains either IP or a DNS address. For file socket it will contain a path starting with `/` which can be used to detect the **file socket mode**. Maybe if we rename **ControlIP** to **ControlAddress** its name would be better suitable; anyway such renaming is a separate thing.
- the only changes needed will be localised in `Connection.cpp`-unit. No changes to other units (Options.cpp, RemoteServer.cpp, BinRpc.cpp, WebServer.cpp). Some error messages printed by program maybe a little misleading (they will mention "port" instead of "file socket") but I think that's acceptable.
- add a couple of lines to `Bind` in `Connection.cpp` to support the file socket mode (instead of new class). This will make the feature transparently usable from other program parts, which don't need to be aware of the new mode at all (they just pass ControlIP to Connection).

And of course you have to make CLI work; otherwise that's a half baked feature. Hopefully a couple of lines in `Connect`-method can make the client-side work.

schnusch:
Yes, you are right, by simply reusing the value of ControlIP it all boils down to changes in `Connection::Bind` and `Connection::DoConnect`. I also added some code to remove the UNIX socket upon closing the listening connection in `Connection::DoDisconnect` and to avoid remote address resolution in `Connection::GetRemoteAddr`.
The CLI client also works. A side-effect that was introduced is that one should now also be able to connect to NNTP servers on local UNIX sockets.

hugbug:
Works perfectly.
Sorry for the delay of merging this. I need to publish v19.1 with fixes and can merge the PR (for v20.0) after that.

PR #417, opened 2017-07-24
made arch validation more robust

stefaang:
The installer fails to parse i686 as a platform on the Google Pixel x686 emulator.
This is caused by some weird behavior of the ValidArch filter on this specific platform.
This fix is less 'spacy' and allows proper platform parsing.

An example of how echo piped into sed works on this platform:

    adb shell
    # var=`echo " aa bb cc " | sed "s/ cc //"`
    # echo $var
    aa bb aa bb cc
    # var=`echo " aa bb cc " | sed "s/ aa //"`
    # echo $var
    aa bb cc

### DIFF
--- a/linux/installer.sh
+++ b/linux/installer.sh
@@ -60,10 +60,11 @@ Error()
 
 ValidArch()
 {
-    LIMARCH=`echo "$ALLARCHS" | grep -w "$1"`
-    if test "$ALLARCHS" = "$LIMARCH"; then
-        return 0
-    fi
+    for arch in $ALLARCHS; do
+        if test $arch = $1; then
+            return 0
+        fi 
+    done
     return 1
 }
 
--- a/linux/installer.sh
+++ b/linux/installer.sh
@@ -60,11 +60,11 @@ Error()
 
 ValidArch()
 {
-    LIMARCH=`echo " $ALLARCHS " | sed "s/ $1 //"`
-    if test " $ALLARCHS " = "$LIMARCH"; then
-        return 1
+    LIMARCH=`echo "$ALLARCHS" | grep -w "$1"`
+    if test "$ALLARCHS" = "$LIMARCH"; then
+        return 0
     fi
-    return 0
+    return 1
 }
 
 PrintArch()
--- a/linux/installer.sh
+++ b/linux/installer.sh
@@ -60,8 +60,8 @@ Error()
 
 ValidArch()
 {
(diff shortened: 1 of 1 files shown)


hugbug:
This unfortunately adds a dependency from "grep" which is a bad thing. There are so many busybox variants with differently limited functionality.

Maybe a loop like this instead?
https://github.com/nzbget/nzbget/blob/develop/linux/installer.sh#L243

hugbug:
Have you tested this? The return statements look wrong.

stefaang:
Yeah. The existing implementation filters for the arch, replaces it by an empty string and checks if ALLARCHS matches the result. If it matches, the arch was actually NOT in ALLARCHS.

The new implementation is a bit more logical, the test returning 0 means it is in the ALLARCHS variable.

hugbug:
Ah, thanks.

The last thing. To make the new code more similar to other one please use upper case for variables and use quotes in test-comparisons. 
```
for TARG in $ALLARCHS; do
   if test "$TARG" = "$1"; then
```
I use TARG here instead of ARCH because ARCH is used in other places and I'm not sure about variables lifetime (if it would be local to the function or would overwrite ARCH used elsewhere).

Thanks.

hugbug:
Great. Thank you!

stefaang:
I found the root cause: android M ships with busybox and toybox.

    root@box:/ # ls -l /system/bin/sed
    lrwxr-xr-x root     shell             2017-09-07 20:44 sed -> toybox
    root@box:/ # echo " aa bb cc " | sed "s/ cc //"
     aa bb aa bb cc
    root@box:/ # echo " aa bb cc " | busybox sed "s/ cc //"
     aa bb

This is a bug in toybox sed ...

stefaang:
And the same for the weird tar behaviour I reported in the android installer: a bug in toybox tar

hugbug:
I guess the installation from Android installer app would work as the included busybox would be [used](https://github.com/nzbget/android/blob/master/app/src/main/jniLibs/armeabi/libdaemon.so#L5).

stefaang:
Sure but the x86 build is broken :) that's why I tried to use the builtin tools.
Using the builtin tools makes the android installer compatible with x86, x64 and the ARM flavours.

hugbug:
> Sure but the x86 build is broken :) that's why I tried to use the builtin tools.

Ah, right. Now I remember.

PR #419, opened 2017-07-25
added "listgroupschanged" RPC method

schnusch:
this method takes a optional string parameter and returns one of the following objects:
    `{"groups": groups, "hash": hash}`
        or
    `{"hash": hash}`
where hash is a string identifing groups and groups is identical to the data returned by a call to "listgroups".
The returned hash can be used in later calls to "listgroupschanged". groups is absent in the returned object if the hash given as a parameter is equal to the new one generated for groups because the data that would be returned is exactly the same as it was returned in a previous call.
An absent parameter in the call is treated as a string of length zero.

Using this function in the WebUI greatly reduces the data transmitted during the periodic polls for updates because the entire result of listgroups is not need to be transmitted if it is unchanged.

### DIFF
--- a/daemon/remote/XmlRpc.cpp
+++ b/daemon/remote/XmlRpc.cpp
@@ -34,6 +34,10 @@
 #include "ScriptConfig.h"
 #include "QueueScript.h"
 #include "CommandScript.h"
+#ifndef DISABLE_PARCHECK
+	#include "par2cmdline.h"
+	#include "md5.h"
+#endif
 
 extern void ExitProc();
 extern void Reload();
@@ -141,6 +145,12 @@ public:
 	virtual void Execute();
 private:
 	const char* DetectStatus(NzbInfo* nzbInfo);
+};
+
+class ListGroupsChangedXmlCommand: public ListGroupsXmlCommand
+{
+public:
+	virtual void Execute();
 };
 
 class EditQueueXmlCommand: public XmlCommand
@@ -605,6 +615,10 @@ std::unique_ptr<XmlCommand> XmlRpcProcessor::CreateCommand(const char* methodNam
 	else if (!strcasecmp(methodName, "listgroups"))
 	{
 		command = std::make_unique<ListGroupsXmlCommand>();
+	}
+	else if (!strcasecmp(methodName, "listgroupschanged"))
+	{
+		command = std::make_unique<ListGroupsChangedXmlCommand>();
 	}
 	else if (!strcasecmp(methodName, "editqueue"))
 	{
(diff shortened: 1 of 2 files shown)


hugbug:
I doubt the performance gain is worth the increased complexity of the API:
- the response is already transferred in compressed form (gzip). I just made a test with 68 nzbs in queue. Results: from raw data 113KB only 8KB were transferred, thanks to compression.
- if you keep web-interface opened (a situation where decreased transfer would help) then you probably have active downloads and in that case the groups change on every request.

schnusch:
I currently have 450ish NZBs which is about 70KiB per request when compressed and even when the WebUI is just open and idling it's sending 70KiB/s perpetually which seems kind of wasteful. The best answer would be long polling anyway but that seems like it would need a complete overhaul of the system.
**Edit:** I also especially dislike the idea of the browser handling useless data which just seems way to common nowadays.

hugbug:
Why do you care about 70 KB/s?
Why do you keep webui open when nzbget isn't downloading?

schnusch:
Because I grew up with 1Mbps "broadband".
I also do not think that there is that much added complexity because it simply wraps the listgroups response. I do believe this to be quite maintainable even though these are famous last words.

Sinnlosvoll:
70KB/s is more than some phone contracts can reach. So if you want to check your status on a throttled connection, it would never be up to date. 

If it can work with less bandwidth, then why make it that way? Good software doesn't make assumptions about the users behavior but tries to work with the minimum amount of resources. IMHO that includes minimal bandwidth.

hugbug:
It's 70KB/s if you have a HUGE queue and if you have set update interval to "every second". If you on mobile you can use a longer update interval. Moreover, this optimisation doesn't help loading the first page and you wouldn't keep the webui open on mobile, if there are no active downloads, right? And if there are active downloads this optimisation doesn't work.

Introducing a new method for public API is a big thing and must be well-founded.

In a more typical case the amount of data is about 1KB and the update interval is adjustable by user anyway.

> Good software doesn't make assumptions about the users behavior

Quite the contrary, writing software is all about use cases. If I don't see a use case behind the code I consider the code unnecessary.

I need at least one user (better more than one) who will profit from the feature. To me it looks like the optimisation was made because it was possible, not because it was necessary. I also showed that this optimisation doesn't work in real life.

---

If there is a demand for decreasing bandwidth of webui, then we can discuss this. In that case I would rather prefer a more general solution applicable to any API method (method "history" comes to mind first). And we should then discuss other bottlenecks such as:
- one ui update involves four API requests. This should be minimised to one.
- nzbget web-server doesn't support keep-alive connections requiring connection negotiations on each request (four times per ui update). That becomes even more problematic if communicating via SSL because it requires TLS handshake, which is a costly operation. If you accessing nzbget on mobile you probably use SSL, right?

For the first problem XML-RPC has multicalls (which nzbget supports). JSON-RPC doesn't have such concept but we could develop a home grown solution for own use. The hash-idea can be used as part of a solution.

Both problems are solvable. The reason they are not solved in nzbget yet is because I didn't see a real gain in putting efforts in this direction. If you now tell me this is important I may reconsider and we can develop a solution together.

schnusch:
I get that introducing API changes is a big deal and I didn't even notice the missing keep-alive because I am running NZBGet behind nginx which fixes that.
The fix was created because I noted the bandwidth and its unnecessity because for a tab hoarder the WebUI is always opened and more often than not NZBGet is idling.
I am also aware that listgroupschanged is a very UI centric method and I am not even sure if the UI should solely rely on those RPC calls because I do think some polling or something similar is a better way to go but alas that is no simple task. This PR is an attempt to go into vaguely that direction without too invasive modifications.

hugbug:
Inspired by your PR and the discussion around it I've started an initiative of improving web-interface performance and efficiency.

Please take a look at #421. I've already implemented two improvements which reduce initial load time of web-interface.

One of the item in todo list is "reduce amount of data transferred on UI updates by avoiding sending of unchanged data". That's what you did in this PR but I want to implement it on another level. Instead of extending API the idea is to do the hash calculation on transport level - in `WebServer.cpp` where the generated page is sent to the client. The hash will be sent as custom header. The client sends the known hash via custom header as well.

Pros:
- no extending of API necessary;
- works with all API methods (currently useful for `listgroups` and `history`, maybe for `status` too).

Cons:
- requires work with custom headers.

hugbug:
Superseded by #422.

Issue #421, opened 2017-07-27
Performance improvements in web-interface


hugbug:
Inspired by discussion in #419.

There is a potential to improve web-interface in regard to performance and efficiency when transferring data between web-browser and NZBGet.

This issue is to investigate areas of web-interface where such improvements are possible and are worth a possibly increased complexity. The following areas were identified so far:
- [x] reduce number of requests when initially loading web-interface;
- [x] call multiple API methods simultaneously;
- [x] reduce amount of data transferred on UI updates by avoiding sending of unchanged data;
- [x] support for keep alive connections.

hugbug:
## Reduce number of requests when initially loading web-interface

Web-interface in NZBGet consists of one html-page (index.html), many javascript-files, two css-files and a couple of image files. After it is loaded the javascript-application communicates with NZBGet via API to get information of current program state.

- **html**: 1 file;
- **javascript**: 4 library files + 11 nzbget files;
- **css**: 1 library file + 1 nzbget file;
- **images**: 4 files.
- **api**: 6 calls.

When loading web-interface the browser has to load each required file with a separate request.
**Total:** 28 requests. 

Javascript files represent a half of all load requests. A known technique to minimise the number of requests is to pack all files into one big file. That can be achieved when building (compiling) NZBGet. This would however slow down the development of web-interface.

Instead of combining files during build step another solution was developed. NZBGet's web-server can  join the files. To instruct the web-server what to join a special syntax for URL requests were introduced.

Instead of:
```html
<script language="javascript" type="text/javascript" src="index.js"></script>
<script language="javascript" type="text/javascript" src="util.js"></script>
<script language="javascript" type="text/javascript" src="downloads.js"></script>
...
```
we now write:
```html
<script language="javascript" type="text/javascript" 
    src="combined.js?index.js+util.js+downloads.js+..."></script>
```

Two css-files were also combined into one.

### What about debugging?
It's not really debugging friendly to have one javascript file consisting of more than 10K lines. To make debugging easier the combining of javascript-files is active only when NZBGet is compiled in release mode. In debug builds each javascript-file is loaded individually.

### Results
The total number of requests during initial loading of web-interface has been reduced from 28 to 14. The amount of transferred data isn't affected much by this change (as expected).

How much is the profit in (milli)seconds? See the next post.

hugbug:
## Call multiple API methods simultaneously
[...]

PR #422, opened 2017-07-29
#421: added support for Etag an If-None-Match HTTP headers

schnusch:
The web server now support Etag generation for static files and some RPC
methods. If If-None-Match is given in the request and matches with the Etag
generated for the response than no data is sent and 304 or 412 is returned.

The JavaScript RPC calls also support the new HTTP error code by buffering
Etags and responses and will reuse the previous response if 412 is returned.

### DIFF
--- a/daemon/remote/WebServer.cpp
+++ b/daemon/remote/WebServer.cpp
@@ -25,6 +25,10 @@
 #include "Options.h"
 #include "Util.h"
 #include "FileSystem.h"
+#ifndef DISABLE_PARCHECK
+	#include "par2cmdline.h"
+	#include "md5.h"
+#endif
 
 static const char* ERR_HTTP_BAD_REQUEST = "400 Bad Request";
 static const char* ERR_HTTP_NOT_FOUND = "404 Not Found";
@@ -174,6 +178,10 @@ void WebProcessor::ParseHeaders()
 		{
 			m_forwardedFor = p + 17;
 		}
+		else if (!strncasecmp(p, "If-None-Match: ", 15))
+		{
+			m_oldetag = p + 15;
+		}
 		else if (*p == '\0')
 		{
 			break;
@@ -314,7 +322,7 @@ void WebProcessor::Dispatch()
 		processor.SetUserAccess((XmlRpcProcessor::EUserAccess)m_userAccess);
 		processor.SetUrl(m_url);
 		processor.Execute();
-		SendBodyResponse(processor.GetResponse(), strlen(processor.GetResponse()), processor.GetContentType());
+		SendBodyResponse(processor.GetResponse(), strlen(processor.GetResponse()), processor.GetContentType(), processor.IsCachable());
 		return;
 	}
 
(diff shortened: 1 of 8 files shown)


hugbug:
**This is fantastic!**

It's cool that the idea of cashing API results can be applied to static files. I didn't knew about ETags.

I've tested this in Chrome, Firefox and Safari (all on desktop). Safari doesn't send ETag for "index.html"; Firefox and Chrome send ETags for all resources including "index.html". I'm going to do more tests on mobile to see how the browsers behave there.

I've made [small adjustments](https://github.com/nzbget/nzbget/commit/75843c700ad0934f691abf810488c25ddea5814a). I've imported the PR and committed there but it looks like the changes are in a separated branch. I guess you can merge or cherry-pick that commit into your branch.

I have two questions:

1. Why do you limit what API-methods can be used with ETag?
I would simply allow all methods and let the client decide with which methods to use caching.

2. Since I didn't knew about ETags I've read the docs and it seems the code "412 Precondition Failed" isn't suitable for the purpose. 

> The HTTP 412 Precondition Failed client error response code indicates that access to the target resource has been denied. This happens with conditional requests on methods other than GET or HEAD when the condition defined by the If-Unmodified-Since or If-None-Match headers is not fulfilled. In that case, the request, usually an upload or a modification of a resource, cannot be made and this error response is sent back.

It seems "412" indicates a real error; browsers also show it as red (in developer tools) which scares.

Wouldn't it be better to use the same "304 Not Modified" (even if it's not supposed to be used with method POST)?

schnusch:
If-None-Match is actually a condition on which to evaluate the request and not to send the response. By limiting the cachability to methods that do not change the state in any way we get around this. If If-None-Match was used with methods that alter the state it would have to be evaluated as a pre-condition.

The 412 code isn't pretty but the RFC [1] says:
> An origin server MUST NOT perform the requested method if the condition evaluates to false; instead, the origin server MUST respond with either a) the 304 (Not Modified) status code if the request method is GET or HEAD or b) the 412 (Precondition Failed) status code for all other request methods.

so it would be non-standard if it returned anything else. One might argue that all those methods that are cachable are equivalent to `safe methods` [2] and it is OK to return 304 anyway.

[1] https://tools.ietf.org/html/rfc7232#section-3.2
[2] https://tools.ietf.org/html/rfc2616#section-9.1

hugbug:
> I've made small adjustments. I've imported the PR and committed there but it looks like the changes are in a separated branch. I guess you can merge or cherry-pick that commit into your branch.

Don't bother with this. I'm merging the PR.

hugbug:
I figured out how to push to your branch and added the commit mentioned above. Please pull on your side. I'm leaving the PR open until we complete the work. It's better this way because if I merge you will need to create a new PR but now we both can push into PR branch.

hugbug:
See how it looks in browser console:

screen shot 2017-07-30 at 11 17 48

With "304" no errors are shown.

I think the idea of "412" is the following: the client sends a change request and it uses etag to indicate to server which version of resource it had before sending the change request. That allows server to validate that the client had the latest version and then apply the changes. If server determines that the client had an older version the serves denies applying the changes to avoid loosing of data. Therefore the code is also in 4xx-range (real errors). That's how I understand it.

hugbug:
NZBGet API already had a concept of safe methods. Safe methods can be called using GET. For example http://127.0.0.1:6790/jsonrpc/listgroups.

I've made all methods which were marked as safe cacheable (pushed into PR branch).

hugbug:
**Another optimisation:** if cached results are to be used the table updates for download list and history are skipped. That saves a lot of javascript work and should show good on CPU usage. Especially useful if you keep the tab opened in background when nzbget is idle.

hugbug:
I'm ready to merge except that I'd like to use "304" for API calls to avoid browser errors in the console.

Are you OK with this? Please do the change or let me know if you want me to do it.

schnusch:
That's fine by me as long as it is never used for unsafe functions.

hugbug:
### Results for initial web interface loading
When loading the page the total transferred data reduced from **294KB** to **36KB** provided the browser has nzbget web-interface in cache. I have no idea how big the chances to have the data in cache.

I hope that change will allow much better loading times for mobile users.

### What about these 36KB?
Two requests: method `config` (7KB) and method `configtemplates` (21KB). Would it be crazy to try to improve this too?

I see two possible ways:

1. Use method GET when fetching these resources. I don't know if the browser can cache XMLHttpRequest responses.

2. Save response into local storage for later use.

Do you think it is worth? That would reduce the total transferred data to about 8KB (if available in cache of course). Would it be too good? 😉 

Would you like to investigate this?

schnusch:
Firefox takes care of caching when a GET XMLHttpRequest is made and returns 200 even if actually a 304 was returned. This seems to have been part of the specifications [1] but I cannot seem to find it in the current WHATWG specs.

Completely going for GET requests when calling safe methods seems to be the right way to go because it would be in spirit with HTTP and would leverage an existing browser functionality instead of implementing it oneself.

[1] https://stackoverflow.com/a/16817752

hugbug:
Agree on GET.

For api methods with parameters they must be passed in correct order, parameter names are ignored and can even be omitted. For example api method "history" has one boolean parameter. It can be passed like this: http://192.168.1.3:6792/jsonrpc/history?=true
For mehod "log" with two parameters: http://192.168.1.3:6792/jsonrpc/log?=0&=10

The RPC method "call" can be extended with another parameter telling  if "GET" or "POST" (default) to use. Maybe it's better to replace rarely used parameter "Timeout" with "Options" which can hold both timeout and http-method.

We still need to know that the data is coming from cache for download list and history to be able to optimize updating behaviour. Also currently RPC reuses previous response as already parsed json object. We should keep this optimization and not parse every time (which would happen if we would rely solely on browser).

When implemented the loading of webui should further speed up. Even downloads and history can be loaded from cache if not changed. Really cool :+1:

Would you like to work on this or should I do?

schnusch:
RPC now keeps a list of safe methods and uses GET requests when one of them is called, otherwise a POST is used as before. The handling of 412 errors and the custom caching was removed.

Maybe Etag generation should be turned off for responses of POST requests in general.

**Edit:** To check if the response was cached Etags but not the responses could be tracked manually again or a Date header could be introduced [1].
[1] https://stackoverflow.com/a/16041467

hugbug:
> To check if the response was cached Etags but not the responses could be tracked manually again 

Yes, please because otherwise the optimized updates of downloads and history don't work.

Another issue is that response json is always parsed even if not used in downloads and history. We could pass a new parameter from there to instruct RPC to not parse response at all.

hugbug:
Merged into **develop**.

I had to make the merge manually to correct commit messages (added missing issue references). Unfortunately GitHub doesn't recognise manually merged PRs as merged and will show it as "closed".

PR #437, opened 2017-08-31
Don't rely on sizes of externally generated files

sinic:
My `abc.00` is smaller than 4704 bytes, breaking `rename/rename_opt2_test.py::test_rename_obf3dmf2` due to a failed `str.replace`.

Matching arbitrary sizes (with `re.sub`) prevents problems like this.


hugbug:
You are right of course. Thanks.

PR #453, opened 2017-09-28
Updated scripts to accept special characters in password

BernCarney:
Fix for #453

### DIFF
--- a/scripts/EMail.py
+++ b/scripts/EMail.py
@@ -169,10 +169,8 @@ if (os.environ.get('NZBPO_STATISTICS') == 'yes' or \
 
 	if host == '0.0.0.0': host = '127.0.0.1'
 
-	# Build an URL for XML-RPC requests
-	d = {'u': username, 'p': password, 'h': host, 'po': port}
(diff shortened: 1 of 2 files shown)


hugbug:
Great. Thank you!

Can you please change it to quote the string when building rpcurl, not when assigning to variable `password` as that would be more logical. Also `username`must be quoted as well.

BernCarney:
Will do.

BernCarney (commented):
Updated PR

hugbug:
Thanks for your contribution!

This dictionary feels like overkill 😄  Why not just use `quote`when building rpcUrl?
```python
rpcUrl = 'http://%s:%s@%s:%s/xmlrpc' % (quote(username), quote(password), host, port);
```

BernCarney:
No problem!

Sometimes the simple solution is the last one you think of :P

hugbug:
Would you like to commit a change for that or should I do it myself?

BernCarney:
@hugbug all set

hugbug:
Thank you!
Merged.

PR #475, opened 2017-11-13
#474: Add NETDB_* defines, musl compilation workaround

### LINKED ISSUE #474: Build fails on musl libc caused by ce84463
musl doesn't define `NETDB_*` constants, as used [here](https://github.com/nzbget/nzbget/blob/ce844367e7e3aa69ba872400ec4cde6c117fa6c6/daemon/connect/Connection.cpp#L1213) and therefore no longer compiles as of ce844367e7e3aa69ba872400ec4cde6c117fa6c6

Build log: https://drone.adam-ant.co.uk/Adam-Ant/docker-nzbget/41/4

[Here is a patch](https://gist.github.com/mbuelte/3666df1345828a807526767eacf4f52e) for the same issue for another piece of software, adding the missing defines

frebib:
Fixes #474 

Added missing NETDB_* defines according to glibc source: https://github.com/bminor/glibc/blob/1ffe1ccb6ec5771765f1f6f0c439ed07bf345d67/resolv/netdb.h#L71-L75

### DIFF
--- a/daemon/connect/Connection.h
+++ b/daemon/connect/Connection.h
@@ -33,6 +33,16 @@
 #include "TlsSocket.h"
 #endif
 
+/*
+ * Fix missing NETDB defines when using musl libc
+ */
+#if !defined(NETDB_INTERNAL)
+#  define NETDB_INTERNAL (-1)
+#endif
+#if !defined(NETDB_SUCCESS)
+#  define NETDB_SUCCESS (0)
+#endif
+
 class Connection
 {
 public:


hugbug:
Thanks for your contribution.

The code parts referencing these constants were taken from Android C lib. Looking at how the functions are used in NZBGet I see that error codes are not evaluated at all. Therefore It's better to rework the code to remove references altogether.

Would you like to that or should I do?

We need to remove last parameter in functions `android_gethostbyname_internal` and `android_read_hostent`.

frebib:
Yes, I agree that removing the references completely is the better solution if they're not used. 
My fix was a quick and dirty fix to fix compilation against musl.

You probably better understand what is going on in the DNS resolution code :)

hugbug:
Fixed with 18d138648ba9de01311746e8a1ecd77e973d8434.

Thanks again for your help!

frebib:
Oh, that was easier than I was expecting.
Thanks for the fix!

PR #497, opened 2018-01-20
LibreSSL support

bket:


hugbug:
Thanks. What about `configure`, doesn't it require adjustments for LibreSSL?

bket:
No, configure does not require any changes. LibreSSL is a fork of OpenSSL with as main focus modernising the codebase. It is backwards-compatible with OpenSSL.

hugbug:
Thank you!

PR #536, opened 2018-05-09
nserv: Show IP address of incoming connection

sanderjo:
Like `[INFO] [1] Incoming connection from: 127.0.0.1`


hugbug:
Thank you.
Was merging on mobile and got commit message screwed a little :(

Issue #567, opened 2018-07-20
AddressSanitizer: heap-buffer-overflow daemon/remote/XmlRpc.cpp:896 in XmlCommand::NextParamAsInt(int*)


sanderjo:
I was _playing_ with Address Sanitizer (https://github.com/google/sanitizers/wiki/AddressSanitizer): compiled nzbget (github version) with it, started nzbget (worked), but as soon as I access nzbget via the webinterface, nzbget stops and the console says the below. So a heap-buffer-overflow daemon/remote/XmlRpc.cpp:896 in XmlCommand::NextParamAsInt(int*).

Is this useful? 

Please note: just _playing_ with ASan.

```
#0 0x7fd8477d9d2f in __interceptor_pthread_create (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x37d2f)
#1 0x5582c0fd1413 in Thread::Start() daemon/util/Thread.cpp:115
#2 0x5582c0c905fa in NZBGet::StartRemoteServer() daemon/main/nzbget.cpp:508
#3 0x5582c0ca10f7 in NZBGet::StartRemoteServer() daemon/main/nzbget.cpp:331
#4 0x5582c0ca10f7 in NZBGet::Run(bool) daemon/main/nzbget.cpp:711
#5 0x5582c0ca1628 in RunMain() daemon/main/nzbget.cpp:1022
#6 0x5582c0bb6840 in main daemon/main/nzbget.cpp:164
#7 0x7fd845b00b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
SUMMARY: AddressSanitizer: heap-buffer-overflow daemon/remote/XmlRpc.cpp:896 in XmlCommand::NextParamAsInt(int*)
                                                                                                                                                                                        
Shadow bytes around the buggy address:
                    
  0x0c067fffebc0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fffebd0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fffebe0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fffebf0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fffec00: fa fa fd fd fd fd fa fa 00 00 03 fa fa fa 00 00
[...]

PR #569, opened 2018-07-30
NextParamAsInt: Stop parsing at end-of-string

sanderjo:
As discussed on https://github.com/nzbget/nzbget/issues/567

Stop parsing at end-of-string. And to avoid a heap-buffer-overflow with AddressSanitizer.


hugbug:
Thank you!

PR #578, opened 2018-11-09
Add Python 3 compatibility

sichapman:
This should allow emails to be sent when using Python 3. Tested with 3.7.1

### DIFF
--- a/scripts/EMail.py
+++ b/scripts/EMail.py
@@ -91,13 +91,14 @@ import os
 import sys
 import datetime
 import smtplib
-from urllib2 import quote
 
 from email.mime.text import MIMEText
-try:
-	from xmlrpclib import ServerProxy # python 2
-except ImportError:
-	from xmlrpc.client import ServerProxy # python 3
+try: # python 2
+	from urllib2 import quote
+	from xmlrpclib import ServerProxy
+except ImportError: # python 3
+	from urllib.parse import quote
+	from xmlrpc.client import ServerProxy
 
 # Exit codes used by NZBGet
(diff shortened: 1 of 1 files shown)


hugbug:
Thanks.
I believe I have tested the script with Python 3. Could it be that `quote` was moved to `urllib.parse` in more recent versions of Python? If that's the case the modified script will fail in older versions of Python 3. Can you please verify that?

sichapman:
I've checked the Python 3.0 docs (https://docs.python.org/3.0/library/urllib.parse.html#urllib.parse.quote) and the 2to3 docs (https://docs.python.org/2/library/2to3.html#2to3fixer-urllib) and various random Google search results like this issue (https://github.com/pgbovine/OnlinePythonTutor/issues/147) so am fairly confident in it applying to all of 3.x. Can't explain your original tests working though - I'm in the middle of a server rebuild so can't test it personally for a few days, but will then if you want.

sichapman:
OK. I've tested with 3.0 - the original script does fail as expected with the error importing urllib2. The script in this PR gets past that but then fails later for a different reason, a base64 error. After checking through release notes, it appears  that was a core Python bug that was fixed in 3.1.3 (https://hg.python.org/cpython/raw-file/v3.1.3/Misc/NEWS issue 4768), and indeed that version works. So it looks like 3.1.3 is going to have to be the minimum Python 3 version for emails even after this PR. Supporting Python 3 versions only after 3.1.3 (8 years old) is better than none at all, I guess!

hugbug:
Thank you.
Would you like to provide a similar patch to Logger.py script?

PR #580, opened 2018-11-17
ArticleDownloader/Decoder: Fix off by one offset

fedux:
When an article has yparts and the "begin" is 0 then the articleOffset
will be negative and the first article will fail to decode.

For example with:

  =ybegin part=1 line=128 size=...
  =ypart begin=0 end=...

, then we now use the beginning position reported by the article part.

### DIFF
--- a/daemon/nntp/Decoder.cpp
+++ b/daemon/nntp/Decoder.cpp
@@ -207,6 +207,7 @@ void Decoder::ProcessYenc(char* buffer, int len)
 		{
 			pb += 7; //=strlen(" begin=")
 			m_beginPos = (int64)atoll(pb);
+			if (!m_beginPos) m_beginPos = 1;
 		}
 		pb = strstr(buffer, " end=");
 		if (pb)
@@ -275,7 +276,7 @@ Decoder::EStatus Decoder::Check()
 	{
 		case efYenc:
 			return CheckYenc();
-			 
+
 		case efUx:
 			return CheckUx();
 


fedux:
Well, this now works for me and is passing the tests but somehow it doesn't feel right. I'll try to make a better fix while also waiting for comments.

hugbug:
Thank you for posting the PR.

> =ypart begin=0

That means we have a malformed article. It should be detected in ArticleDownloader and reported as failed which is a correct behaviour.

If the posting software improperly counts offsets from 0 this will affect all the segments, not only the first one. Even if we can detect and correct the error for the first segment, we can't detect errors for all other segments and the download will fail anyway.

Do you have a real life nzb which shows this kind of problem?

fedux:
Yeah, I think this is tricky: On one hand I think you are right and the article is "malformed" (although is a bit unclear from the spec if it can or cannot start from zero, but the example in yenc.org starts at 1); on the other hand other clients download it fine.

I have a local setup with innd as a server and pan/0.145 as the client and uploading a file from it gives these ranges:

```
=ybegin part=1 line=128 size=6730796 name=How-to-open-a-wine-bottle.mp4
=ypart begin=0 end=524288
=yend size=524288 part=1 pcrc32=54157b34

=ybegin part=2 line=128 size=6730796 name=How-to-open-a-wine-bottle.mp4
=ypart begin=524288 end=1048576
=yend size=524288 part=2 pcrc32=a2b618b5

=ybegin part=3 line=128 size=6730796 name=How-to-open-a-wine-bottle.mp4
=ypart begin=1048576 end=1572864
=yend size=524288 part=3 pcrc32=1591d5a6

...
```
, which not only starts from 0 but the next part "begin" is equal to the previous part "end" which is also wrong and then also the size doesn't match. I raised a PR to pan fixing that and when merged it will start from 1 and the next part will be correct.

Also, if nzbget wants to be strict it should also reject the other parts as the size doesn't match if we also count the starting byte: `1048576 - 524288 + 1 != 524288`

Pan and sabnzb are both able to download the file correctly tho.

fedux:
@hugbug, as this is  clearly a problem with a malformed article I don't mind if we just close the PR. If you think it's worth it I can create an issue to follow up on this otherwise we just close it.

hugbug:
Yes, I think it's better to close.
Thanks for your contribution though!

PR #590, opened 2018-12-21
Fix compile warnings

fedux:
This fixes many compile warnings produced when compiling the code with `-Wall`. For the explanation and purpose of the warning see the compiler man page.


hugbug:
Thanks for your contribution. A description would be helpful.

From all the changes I think I'm going to accept these two: 
- sign-compare;
- unused-variable.

For other changes:
- format - makes code unnecessary uglier;
- reorder - I don't know what the purposes of changes were (a description of your intent would be helpful);
- nonnull - fixes the warning for you but produces a warning in other compilers (dereference of null pointer);
- maybe-uninitialized - false diagnostic;
- misleading-indentation and class-memaccess - code in third party libraries. I'd rather not change it;
- stringop-truncation/stringop-overflow - introduces usage of `strcpy` which other compilers consider  dangerous.

fedux:
Hi @hugbug , for the reason of each of the fixes you can check the man page of gcc for example for a detailed explanation.

For these other changes:

 - format - makes code unnecessary uglier;
   The reason is that to support both 32bits and 64bits platforms the proper format for printf and similar depends on the length of the word and thus it needs to be defined in macros.

 - reorder - I don't know what the purposes of changes were (a description of your intent would be helpful);
  From the man page: 
```
       -Wreorder (C++ and Objective-C++ only)
           Warn when the order of member initializers given in the code does not match the order in which they must be executed.  For instance:

                   struct A {
                     int i;
                     int j;
                     A(): j (0), i (1) { }
                   };

           The compiler rearranges the member initializers for "i" and "j" to match the declaration order of the members, emitting a warning to that effect.
```


 - nonnull - fixes the warning for you but produces a warning in other compilers (dereference of null pointer);
   Ok, which compiler? Did it produce a warning before?

 - maybe-uninitialized - false diagnostic;
 Are you sure? This is the context:

```
    unsigned int codepoint;
    while (*utfstr != 0)
    {
        unsigned char ch = (unsigned char)*utfstr;
        if (ch <= 0x7f)
            codepoint = ch;
        else if (ch <= 0xbf)
            codepoint = (codepoint << 6) | (ch & 0x3f);
        else if (ch <= 0xdf)
            codepoint = ch & 0x1f;
```
, if `ch` doesn't match the first `if` but matches the first `else if`  then `codepoint` will be used uninitialized.
 
 - misleading-indentation and class-memaccess - code in third party libraries. I'd rather not change it;
   OK.

 - stringop-truncation/stringop-overflow - introduces usage of strcpy which other compilers consider dangerous.
  Then problem is that the way the original `strncpy` is used is also used in an unsafe way because it checks the length of the `source` buffer instead of the `destination` buffer causing a posible overflow. Using `strcpy` instead doesn't make it any worse as it has the same effect as the current wrong use of `strncpy` and the `destination` buffer is already pre-allocated to the right size a few lines above. I could split this commit in two to address both warnings separately.

The whole purpose of these fixes is to be able to compile with `-Wall -Werror` and enforce better code quality.

hugbug:
> maybe-uninitialized - false diagnostic;
Are you sure? This is the context:
 
You are right. That's a bug. Good thing to have this warning enabled then.

> nonnull - fixes the warning for you but produces a warning in other compilers (dereference of null pointer);
Ok, which compiler? Did it produce a warning before?

That code is 10 years old, I don't remember what it was back then. Right now Clang produces warning when compiling in analyse mode. Not a big issue probably if the purpose is -Wall. I can accept the change.

> stringop-truncation/stringop-overflow

Can you change the code somehow to not use `strcpy`?

You convinced me, I'm going to accept the whole PR. Although I don't like some changes they are probably for good.

fedux:
> Can you change the code somehow to not use `strcpy`?

Changed. If you think there is something else in this PR that could be improved let me know.

hugbug:
Thank you.

Issue #351, opened 2017-03-19
Decrease idle cpu/power usage


Mike1445:
When idling nzbget seems to still use quite a bit of cpu. Looking at powertop it is by far the most consuming process on my machine.

powertop:
Usage       Events/s    Category       Description
1.8 ms/s     **212.3**        Process        /usr/bin/nzbget -c /etc/nzbget/nzbget.conf -D

top shows around 0.3% cpu usage:

top:
 PID  USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
 5451 nzbget   20   0  823788  13184   4880 S   **0.3**  0.3   0:10.04 nzbget

strace shows only sleep nanosleep system calls every 100 milliseconds:

strace:
restart_syscall(<... resuming interrupted nanosleep ...>) = 0
nanosleep({0, 100000000}, NULL)         = 0
nanosleep({0, 100000000}, NULL)         = 0
nanosleep({0, 100000000}, NULL)         = 0

Maybe increasing the hardcoded timeout from 100 m.s. to 1 second will help?

```
		// sleep longer in StandBy
		int sleepInterval = downloadStarted ? 0 : standBy ? **100** : 5;
		usleep(sleepInterval * 1000);
```

hugbug:
Longer wait intervals may reduce responsiveness.
Nonetheless, why don't you try different values and tell us how it performs?

RollingStar:
Here's the file if anyone wants to investigate.

https://github.com/nzbget/nzbget/blob/62b3d47b43d91109586b352c22844146c495c4dc/daemon/queue/QueueCoordinator.cpp#L231

hazcod:
So we would only reduce responsiveness (so start of a download) whenever nothing is downloading?
Does this include the web thread as wel?

In any case, upping this to 500ms shouldn't be too hard of an impact.
Can you try that @Mike1445 ?

vdboor:
What does nzbget need to monitor when nothing is downloading? Web requests can block on the socket, queues might wait for fnotify. Hopefully nothing needs to be polled all the time.

fedux:
I found the same problem today (high CPU usage) when trying to optimize my power usage. I agree with @vdboor and most loops could be changed to event based instead of polling. Is there any interest in this to happen? I'd try to do this myself in my spare time and submit the patches if so.

hugbug:
[...]

PR #593, opened 2018-12-30
Idle CPU usage improvements

fedux:
This PRs simplifies mutex and threads, fixes some unprotected members and includes a few changes to improve CPU usage when idle.

Part of a solution to https://github.com/nzbget/nzbget/issues/351

### DIFF
--- a/daemon/feed/FeedCoordinator.h
+++ b/daemon/feed/FeedCoordinator.h
@@ -109,6 +109,8 @@ private:
 	bool m_force = false;
 	bool m_save = false;
 	FeedCache m_feedCache;
+	Mutex m_pauseMutex;
+	std::condition_variable m_pauseCV;
 
 	void StartFeedDownload(FeedInfo* feedInfo, bool force);
 	void FeedCompleted(FeedDownloader* feedDownloader);
--- a/daemon/postprocess/PrePostProcessor.cpp
+++ b/daemon/postprocess/PrePostProcessor.cpp
@@ -59,13 +59,25 @@ void PrePostProcessor::Run()
 
 	while (!IsStopped())
 	{
-		if (!g_Options->GetTempPausePostprocess() && m_queuedJobs)
+		if (g_Options->GetTempPausePostprocess())
+		{
+			// Postprocess is paused: just wait and loop
+			usleep(200 * 1000);
+			continue;
+		}
+
+		if (m_queuedJobs)
 		{
 			// check post-queue every 200 msec
 			CheckPostQueue();
+			usleep(200 * 1000);
 		}
-
-		usleep(200 * 1000);
+		else
+		{
+			// Wait until we get the stop signal or more jobs in the queue
+			UniqueLock lk(m_pauseMutex);
+			m_pauseCV.wait(lk, [&]{ return IsStopped() || m_queuedJobs; });
+		}
 	}
 
 	WaitJobs();
@@ -159,6 +171,9 @@ void PrePostProcessor::Stop()
 			((DirectUnpack*)nzbInfo->GetUnpackThread())->Stop(downloadQueue, nzbInfo);
 		}
 	}
+
+	// Trigger the stop signal
+	m_pauseCV.notify_all();
 }
 
 /**
@@ -341,6 +356,9 @@ void PrePostProcessor::NzbDownloaded(DownloadQueue* downloadQueue, NzbInfo* nzbI
 		}
 
 		downloadQueue->Save();
+
+		// We have more jobs in the queue, notify Run()
+		m_pauseCV.notify_all();
 	}
 	else
 	{
--- a/daemon/postprocess/PrePostProcessor.h
+++ b/daemon/postprocess/PrePostProcessor.h
@@ -42,8 +42,10 @@ protected:
 	virtual void Update(Subject* caller, void* aspect) { DownloadQueueUpdate(aspect); }
 
 private:
-	int m_queuedJobs = 0;
+	std::atomic_int m_queuedJobs{0};
 	RawNzbList m_activeJobs;
+	Mutex m_pauseMutex;
+	std::condition_variable m_pauseCV;
 
 	void CheckPostQueue();
 	void CheckRequestPar(DownloadQueue* downloadQueue);
--- a/daemon/queue/DownloadInfo.h
+++ b/daemon/queue/DownloadInfo.h
@@ -171,7 +171,7 @@ public:
 	void SetCompletedArticles(int completedArticles) { m_completedArticles = completedArticles; }
 	bool GetParFile() { return m_parFile; }
 	void SetParFile(bool parFile) { m_parFile = parFile; }
-	Guard GuardOutputFile() { return Guard(m_outputFileMutex); }
+	UniqueLock GuardOutputFile() { return UniqueLock(*m_outputFileMutex); }
 	const char* GetOutputFilename() { return m_outputFilename; }
 	void SetOutputFilename(const char* outputFilename) { m_outputFilename = outputFilename; }
 	bool GetOutputInitialized() { return m_outputInitialized; }
--- a/daemon/util/Thread.cpp
+++ b/daemon/util/Thread.cpp
@@ -23,14 +23,7 @@
 #include "Thread.h"
 
 int Thread::m_threadCount = 1; // take the main program thread into account
-std::unique_ptr<Mutex> Thread::m_threadMutex;
-
-void Thread::Init()
-{
-	debug("Initializing global thread data");
-
-	m_threadMutex = std::make_unique<Mutex>();
-}
+Mutex Thread::m_threadMutex;
 
 Thread::Thread()
 {
--- a/daemon/util/Thread.cpp
+++ b/daemon/util/Thread.cpp
@@ -25,52 +25,6 @@
 int Thread::m_threadCount = 1; // take the main program thread into account
 std::unique_ptr<Mutex> Thread::m_threadMutex;
 
-
-Mutex::Mutex()
-{
-#ifdef WIN32
-	InitializeCriticalSection(&m_mutexObj);
-#else
-	pthread_mutex_init(&m_mutexObj, nullptr);
-#endif
-}
-
-Mutex::~Mutex()
-{
-#ifdef WIN32
-	DeleteCriticalSection(&m_mutexObj);
-#else
-	pthread_mutex_destroy(&m_mutexObj);
-#endif
-}
-
-void Mutex::Lock()
-{
-#ifdef WIN32
-	EnterCriticalSection(&m_mutexObj);
-#ifdef DEBUG
-	// CriticalSections on Windows can be locked many times from the same thread,
-	// but we do not want this and must treat such situations as errors and detect them.
-	if (m_mutexObj.RecursionCount > 1)
-	{
-		error("Internal program error: inconsistent thread-lock detected");
-	}
-#endif
-#else
-	pthread_mutex_lock(&m_mutexObj);
-#endif
-}
-
-void Mutex::Unlock()
-{
-#ifdef WIN32
-	LeaveCriticalSection(&m_mutexObj);
-#else
-	pthread_mutex_unlock(&m_mutexObj);
-#endif
-}
-
-
 void Thread::Init()
 {
 	debug("Initializing global thread data");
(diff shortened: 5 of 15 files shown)


hugbug:
That's a brilliant work. I have some questions which I'll post inline.

daemon/queue/DownloadInfo.h
```diff
@@ -226,7 +226,7 @@ class FileInfo
 	int m_completedArticles = 0;
 	bool m_outputInitialized = false;
 	CString m_outputFilename;
-	std::unique_ptr<Mutex> m_outputFileMutex;
+	Mutex m_outputFileMutex;
```
hugbug: That creates a mutex for each file (for example rar-file) in download queue. One nzb-item may contain hundreds of files. With a hundred of nzb-files in queue many thousands of file-objects may exist. Are you sure it is OK to have thousands of mutexex created?

The previous implementation of Mutex were using system objects and I didn't want to create too many of system objects. How lightweight std::mutex is? Is there any part of C++ specification saying the mutexes are lightweight enough to be created in thousands without worries?

daemon/util/Thread.cpp
```diff
@@ -41,29 +42,22 @@ void Thread::Start()
...
-	pthread_attr_setdetachstate(&m_attr, PTHREAD_CREATE_DETACHED);
-	pthread_attr_setinheritsched(&m_attr, PTHREAD_INHERIT_SCHED);
-	m_running = !pthread_create(&m_threadObj, &m_attr, Thread::thread_handler, (void *) this);
-	pthread_attr_destroy(&m_attr);
-#endif
+	// wait for the ready signal
+	futureReady.wait();
+
+	// save the native handler to be able to cancel (Kill) the thread and then detach
+	m_threadObj = t.native_handle();
```
hugbug, line 59: At this moment the thread may be already terminated (if it was short-lived). If it was auto-destroying (`m_autoDestroy`) then `this` is already destroyed and we can't access `m_threadObj` anymore.

Would moving this line into thread's lambda function fix the issue?

----

> // save the native handler

A typo: handler -> handle.

hugbug: What is the advantage of using promise/future over a mutex?

Promise/future is a new thing to me and I'm trying to understand how they work and if they are better than manually managed mutexes. 

Promise/future has some kind of thread synchronisation under the hood as well, right? The old code was using one synchronisation (via mutex) to start the thread and increase `m_threadCount`. The new implementation has two synchronisations - one via promise/future and another one via atomic. Generally more synchronisations mean more CPU cache flushes and worse performance.

daemon/queue/DownloadInfo.h
fedux: Yeah, there is no problem with having many std::mutex instances. Just as a test I created a few millions without any problems. AFAIK the C++ specification doesn't deal with the implementation details.
I was also wondering why you have to create the FileInfo objects for Nzb files not active in the queue, any even maybe we can get rid of this mutex completely later.
In any case, if you still prefer to skip this commit then I can do that.

daemon/util/Thread.cpp
fedux: > At this moment the thread may be already terminated (if it was short-lived). If it was auto-destroying (`m_autoDestroy`) then `this` is already destroyed and we can't access `m_threadObj` anymore.

OK, I see what you mean. I think for now I can keep the old mutex but use it only for ensuring this. I wonder if the need for the `m_autoDestroy` can be avoided but I still haven't looked at that part of the code yet.

> What is the advantage of using promise/future over a mutex?

It's higher level and let's the underlying implementation to use the best method to implement it. promise/future are useful to send results and states between threads and are meant to be used only once, which fits nice into this use case.

> The new implementation has two synchronisations - one via promise/future and another one via atomic. 

Using atomic is nice if the thread contention is low as it doesn't stop the threads (it locks but doesn't stop the waiting thread and switch context). In the previous implementation we had a few shared values not protected and using atomic is in my opinion a nice solution as it's more efficient than a mutex (mutexes use atomic operations internally) and thread contention should be very low for this cases.

daemon/queue/DownloadInfo.h
hugbug: > if you still prefer to skip this commit then I can do that.

Yes please. I'm not comfortable with thousands of (unused) mutexes hanging around.

fedux: Changed.

daemon/util/Thread.h
```diff
@@ -73,15 +74,15 @@ class Thread
...
 	HANDLE m_threadObj = 0;
 #else
 	pthread_t m_threadObj = 0;
 #endif
-	bool m_running = false;
-	bool m_stopped = false;
-	bool m_autoDestroy = false;
+	std::atomic_bool m_running{false};
+	std::atomic_bool m_stopped{false};
+	std::atomic_bool m_autoDestroy{false};
```
hugbug: `m_autoDestroy` is set before the thread is started and is never changed later; no need to protect.

As for `m_running` and `m_stopped` - is protecting here really necessary? The thread worker functions usually often check thread status via `IsStopped` to decide if the processing must be aborted. For example in article downloader it is checked after every (small) chunk of downloaded data - that maybe thousands times per second (in every download thread).

I thought the synchronisation can be omitted here since we work with a primitive type (bool) which is read/written with a single CPU operation (hopefully, yeah). Worst case scenario - the thread which reads the value may get the older value of the variable but that's not critical as it will eventually get the correct value somewhat later. Is this asking for troubles or a legitimate optimisation?

fedux: Ok, I will revert `m_autoDestroy` to bool. About the other two, in my opinion it's better to program safe than try to optimize something that doesn't really make a difference in performance, but it's your call if you want this to be reverted as well.

hugbug: > try to optimize something that doesn't really make a difference in performance

Performance was my concern exactly. In my short performance test on x86_64 the performance hasn't changed between before-PR and with-PR. I haven't tested on ARM as these tests are more difficult to perform there. I may do tests later though. I can't test other CPUs (MIPS, PowerPC) at all.

For now I suggest to use atomic (for `m_running` and ´m_stopped´).

daemon/util/Thread.cpp
hugbug: > OK, I see what you mean. I think for now I can keep the old mutex but use it only for ensuring this.

Once this is fixed I can merge the PR.

I've tried this and it worked:

```C++
void Thread::Start()
{
	debug("Starting Thread");

	m_running = true;

	// NOTE: "m_threadMutex" ensures that "t" lives until the very end of the function
	Guard guard (m_threadMutex);

	// start the new thread
	std::thread t([&]{
		{
			// trying to lock "m_threadMutex", this will wait until function "Start()" is completed
			// and "t" is detached.
			Guard guard(m_threadMutex);
		}

		thread_handler();
	});

	// save the native handle to be able to cancel (Kill) the thread and then detach
	m_threadObj = t.native_handle();
	t.detach();
}
```

I've also tried to solve this with promise/future but I needed two pairs of promise/future. The solution with mutex is simpler.

daemon/feed/FeedCoordinator.h
```diff
@@ -31,6 +31,7 @@
 #include "Observer.h"
 #include "Util.h"
 
+#include <condition_variable>
```
hugbug: For this and other places: please move all new includes into [nzbget.h](https://github.com/nzbget/nzbget/blob/develop/daemon/main/nzbget.h#L225).

This improves compiling times with precompiled headers.

fedux: done

daemon/util/Thread.cpp
fedux: Updated

hugbug:
Thanks for applying the changes and a special thank for updating the commits instead of adding new commits with small corrections. I like the clean history.

I have one more small thing - will post an inline comment.

daemon/postprocess/PrePostProcessor.cpp
```diff
@@ -59,13 +59,25 @@ void PrePostProcessor::Run()
...
+		if (m_queuedJobs)
 		{
 			// check post-queue every 200 msec
 			CheckPostQueue();
+			usleep(200 * 1000);
+		}
+		else
+		{
+			// Wait until we get the stop signal or more jobs in the queue
+			std::unique_lock<std::mutex> lk(m_pauseMutex);
```
hugbug: Should probably be:

```C++
UniqueLock lk(m_pauseMutex);
```

Maybe also better to use name `guard` just like in all other places? Not sure about that though.

Consequently `m_pauseMutex` should be declared as `Mutex` instead of `std::mutex`.

And similarly in `FeedCoordinator.cpp/.h`.

daemon/postprocess/PrePostProcessor.h
```diff
@@ -42,8 +42,10 @@ class PrePostProcessor : public Thread, public Observer
 	virtual void Update(Subject* caller, void* aspect) { DownloadQueueUpdate(aspect); }
 
 private:
-	int m_queuedJobs = 0;
+	std::atomic_uint m_queuedJobs{0};
```
hugbug: I use `int` whenever possible. Unsigned types considered dangerous.

Please replace with `std::atomic_int`. There are also few other places with `std::atomic_uint` in other units which should be replaced too. Thanks.

daemon/postprocess/PrePostProcessor.cpp
fedux: > Maybe also better to use name guard just like in all other places? Not sure about that though.

Yeah, I thought about that but as the type is std::unique_lock lock I left it like that because it better matches the real type.

Changed the rest.

daemon/postprocess/PrePostProcessor.h
fedux: changed

hugbug:
I've run into an issue when compiling for linux installer package (cross compiling for many CPU architectures).

The compilation for **armel** breaks with:
```
daemon/main/nzbget.cpp:226:21: error: field ‘m_stopSignal’ has incomplete type ‘std::promise<void>’
  std::promise<void> m_stopSignal;
                     ^
In file included from daemon/main/nzbget.h:229:0,
                 from daemon/main/nzbget.cpp:22:
/mnt/build/toolchain/armel/output/host/usr/arm-buildroot-linux-uclibcgnueabi/include/c++/5.2.0/future:124:11: note: declaration of ‘class std::promise<void>’
     class promise;
           ^
```

After searching found [this explanation](https://lists.debian.org/debian-arm/2013/12/msg00000.html). In short (read the link for details):

> any C++11 program linked to libstdc++ which uses future facilities or transfer exceptions will fail to compile on armel.

Do you have an alternative solution without promise/future maybe? If you don't I would need to `ifdef` promise/future and need to extend the configure-script with detection of promise/future-support.

hugbug:
I think we can use `std::condition_variable` here. I'll try this as an exercise.

fedux:
Yes, it can be replaced by condition_variable, but it's a shame that it doesn't work for armel because that means we won't be able to use async or packed_task either probably. I'll see what I can find about it.

fedux:
Apparently it's already fixed in gcc https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64735

Which compiler version is failing?

hugbug:
It's GCC 5.2.

Even though the problem seems to be fixed in GCC 7.0 I cannot easily drop support of older GCC versions. Armel is an important platform. It was a pain for some users when I migrated to C++11/14 with requirement for GCC 4.9. Even GCC 4.8 supported (but only when compiling in release mode) and I intentionally don't use some C++ features not available on GCC 4.8 in order to maintain compatibility.

Changing system requirements of NZBGet to higher GCC version is possible but it has to bring a big advantage to the project. A simple "there is one place where we could use a new compiler feature" isn't such case. The switch from C++98 to C++11/14 was such case.

If promise/future or async or packed_task will allow to for example increase performance to 10% then the switch may be worth it. If these features just allow to save a couple of lines of code then no.

Maintaining compatibility with many platforms is indeed painful and has costs.

fedux:
debian has backported it to gcc-6 but I see your point. It's a shame because it's not really a requirement of a higher c++ standard but a bug in gcc.

In any case, it's not a big deal to use condition_variable instead.

PR #596, opened 2019-01-04
CPU usage when Idle improvements

fedux:
Part of #351

### DIFF
--- a/daemon/queue/UrlCoordinator.cpp
+++ b/daemon/queue/UrlCoordinator.cpp
@@ -82,8 +82,6 @@ void UrlCoordinator::Run()
 		usleep(20 * 1000);
 	}
 
-	int resetCounter = 0;
-
 	while (!IsStopped())
 	{
 		bool downloadStarted = false;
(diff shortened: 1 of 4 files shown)


daemon/queue/UrlCoordinator.cpp
```diff
@@ -82,38 +82,32 @@ void UrlCoordinator::Run()
...
 	}
 
-	int resetCounter = 0;
-
 	while (!IsStopped())
 	{
 		bool downloadStarted = false;
-		if (!g_Options->GetPauseDownload() || g_Options->GetUrlForce())
+		if ((!g_Options->GetPauseDownload() || g_Options->GetUrlForce()) &&
+			((int)m_activeDownloads.size() < g_Options->GetUrlConnections()))
```
hugbug: `m_activeDownloads.size()` is no longer protected by `DownloadQueue::Guard()`.

hugbug (commented):
Now it takes up to whole second between adding new URL to queue and starting download. Previously it was up to 100ms. Also when there are many URLs in queue it can be up to one second between one URL is finished and another one is started.

hugbug:
Better but the second part isn't fixed yet:

> when there are many URLs in queue it can be up to one second between one URL is finished and another one is started.

daemon/queue/UrlCoordinator.cpp
```diff
@@ -105,15 +103,13 @@ void UrlCoordinator::Run()
...
-		resetCounter += sleepInterval;
-		if (resetCounter >= 1000)
+		if (!downloadStarted && !IsStopped())
 		{
 			// this code should not be called too often, once per second is OK
 			ResetHangingDownloads();
-			resetCounter = 0;
+
+			UniqueLock lk(m_pauseMutex);
+			m_pauseCV.wait_for(lk, std::chrono::seconds(1), [&]{ return IsStopped() || m_hasMoreJobs; });
```
hugbug, line 112: Very high CPU usage (full load of one core) if the program was started with paused queue:

```shell
nzbget -s -P
```

This is caused by `m_hasMoreJobs` initialised to `true` in constructor; it's never recomputed above if download is paused.

Originally `m_hasMoreJobs` was used to inform main class (`nzbget.cpp`) that there is still job to do. It was set to `true` in constructor to make sure the main class doesn't stop preliminary before UrlCoordinator has checked the queue.

PR #600, opened 2019-01-08
TlsSocket: Fix deprecated OpenSSL calls

fedux:
Since OpenSSL 1.1.0 we have:

 - ERR_remove_thread_state, ERR_remove_state: "They are now deprecated
   and do nothing".

 - ASN1_STRING_data: "This function is deprecated: applications should
   use ASN1_STRING_get0_data() instead".

### DIFF
--- a/daemon/connect/TlsSocket.cpp
+++ b/daemon/connect/TlsSocket.cpp
@@ -195,7 +195,9 @@ void TlsSocket::Final()
 	CRYPTO_set_locking_callback(nullptr);
 	CRYPTO_set_id_callback(nullptr);
 #endif
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 	ERR_remove_state(0);
+#endif
 #if	OPENSSL_VERSION_NUMBER >= 0x10002000L && ! defined (LIBRESSL_VERSION_NUMBER)
 	SSL_COMP_free_compression_methods();
 #endif
@@ -214,7 +216,9 @@ TlsSocket::~TlsSocket()
 	Close();
 
 #ifdef HAVE_OPENSSL
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 	ERR_remove_state(0);
+#endif
 #endif
 }
 
@@ -559,7 +563,7 @@ bool TlsSocket::ValidateCert()
 	// hostname verification
 	if (!m_host.Empty() && X509_check_host(cert, m_host, m_host.Length(), 0, nullptr) != 1)
 	{
-		char* certHost = nullptr;
+		const unsigned char* certHost = nullptr;
         // Find the position of the CN field in the Subject field of the certificate
         int common_name_loc = X509_NAME_get_index_by_NID(X509_get_subject_name(cert), NID_commonName, -1);
         if (common_name_loc >= 0)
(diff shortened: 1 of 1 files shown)


hugbug:
Since the old code works fine in 1.1.0 (despite deprecation) what do these changes achieve?
Elimination of warnings?

fedux:
Yes:
```
daemon/connect/TlsSocket.cpp:217:20: warning: ‘void ERR_remove_state(long unsigned int)’ is deprecated [-Wdeprecated-declarations]
  ERR_remove_state(0);
                    ^
In file included from /usr/include/openssl/e_os2.h:13,
                 from /usr/include/openssl/ssl.h:15,
                 from ./daemon/main/nzbget.h:259,
                 from daemon/connect/TlsSocket.cpp:21:
/usr/include/openssl/err.h:260:1: note: declared here
 DEPRECATEDIN_1_0_0(void ERR_remove_state(unsigned long pid))
 ^~~~~~~~~~~~~~~~~~
daemon/connect/TlsSocket.cpp: In member function ‘bool TlsSocket::ValidateCert()’:
daemon/connect/TlsSocket.cpp:575:57: warning: ‘unsigned char* ASN1_STRING_data(ASN1_STRING*)’ is deprecated [-Wdeprecated-declarations]
      certHost = (char*)ASN1_STRING_data(common_name_asn1);
                                                         ^
In file included from /usr/include/openssl/e_os2.h:13,
                 from /usr/include/openssl/ssl.h:15,
                 from ./daemon/main/nzbget.h:259,
                 from daemon/connect/TlsSocket.cpp:21:
/usr/include/openssl/asn1.h:554:1: note: declared here
 DEPRECATEDIN_1_1_0(unsigned char *ASN1_STRING_data(ASN1_STRING *x))
 ^~~~~~~~~~~~~~~~~~
```

fedux:
Documentation:
 https://www.openssl.org/docs/man1.1.0/crypto/ERR_remove_state.html
 https://www.openssl.org/docs/man1.1.0/crypto/ASN1_STRING_data.html

hugbug:
This PR has exposed a memory leak in NZBGet.

Apparently `ERR_remove_state` must be called from each thread which has used OpenSSL. However in NZBGet the function is called only when closing connection. Threads in NZBGet are short lived. One thread opens connection, downloads first article and terminates. Then another thread is created and it reuses the already active connection. Many many threads later the connection is closed and `ERR_remove_state` is executed, once (per connection).

If during download OpenSSL happens to create error queue in each thread then all of these queues remain unfreed.

To fix the issue every thread which might have to do anything with connections (certainly every download thread at least) must call that function.

I've found [that topic](https://curl.haxx.se/mail/lib-2016-08/0137.html) where authors of **libcurl** encountered that issue as well:

> Issue was filed yesterday, identifying yet another memory leak with libcurl using OpenSSL. I suppose it says something when we still after 17 years of working with the OpenSSL API keep finding ourselves leak memory by not understanding how to clean it up properly...

Well said, ha-ha.

I don't know if I should fix that or not. There is no simple and elegant fix for this. I'm certainly not going to put OpenSSL call into common thread termination code in `Thread.cpp` but I also don't want to call it from many other places which deal with connections.

fedux:
On the bright side, there is no leak if you use OpenSSL >= 1.1.0 ;)

"They are now deprecated and do nothing, as the OpenSSL libraries now normally do all thread initialisation and deinitialisation automatically."

And from the 1.0.2 man page:

"ERR_remove_state() is available in all versions of SSLeay and OpenSSL. It was deprecated in OpenSSL 1.0.0 when ERR_remove_thread_state was introduced and thread IDs were introduced to identify threads instead of 'unsigned long'."

So, maybe `ERR_remove_thread_state` would help if openssl < 1.1.0 ?

hugbug:
`ERR_remove_thread_state(nullptr)` is equivalent to `ERR_remove_state(0)` - clears error queue of current thread. Challenges are the same - to find good place to call one of these functions.

Maybe I'll just pretend the issue doesn't exist 😉

fedux:
nothing to see here.... next!

PR #631, opened 2019-05-01
Add support for http header

pduveau:
This feature allows nzbget to delegate authentication to a reverse proxy through a localhost's socket and an http header.


hugbug:
Thank you. Can you please explain why is this necessary.

Can't you use header `Authorization: Basic XXXXXX` (with encoded nzbget username and password) instead?

Or maybe disabling authentication in NZBGet completely if it's handled by proxy server?

pduveau:
This allows to use SSL with client's certficate required (strong authentication) and use certificate subject attribute as te user name without to disable user/password as a fallthrough authentication method. 

The `Authorization: Basic XXXXXX` can not be easely used in reverse proxy conf as the password has not to be known by the proxy server.

I also PR a doc update to explain how to implement that with Apache/Tls/Client Certificate.

hugbug:
Thanks for explanation.

I'm currently preparing v21 stable release and will merge this PR afterwards.

PR #667, opened 2019-11-03
Lalufu/feature libcurl support

Lalufu:
Add libcurl support for HTTP(S) downloads
    
    Motivation
    ==========
    
    This introduces support for using libcurl to handle HTTP(S) downloads
    instead of the built-in TLS and HTTP client. The main driver for this is
    poor handling for some HTTP(S) servers from the internal code. A
    specific example is api.nzbgeek.info, which at the time of this patch
    does not work with nzbget over HTTPS.
    
    The web server at api.nzbgeek.info reacts to a SNI (server name indication)
    sent by the HTTPS client with a "The server name sent was not
    recognized" TLS warning message during the handshake. This is a warning
    message, and it indicates that the web server is not configured
    properly, but it is not necessarily fatal. The web server then proceeds
    to present a valid certificate for the api.nzbgeek.info host name. All
    browsers and other download tools (like curl and wget) do not treat this
    warning as fatal, and continue the HTTPS handshake.
    
    At least when compiled with GnuTLS as the SSL engine nzbget treats the
    warning as fatal, and aborts the connect, making it impossible to
    download files via HTTPS from api.nzbgeek.info.
    
    Patching the TLS code path in nzbget to not treat this error as fatal is
    not overly difficult, but once that is done the download does still not
    succeed. This time the content is downloaded, but the interaction
    between the HTTP client and the SSL layer for tearing down the
    connection is slightly flawed, leading to another SSL alert ("TLS
    session not properly terminate"), which is also treated as fatal.
    
    At this point it seemed prudent to investigate offloading the whole
    HTTP(S) handling to a dedicated and tested library.

    Changes
    =======
    
    libcurl support is added as optional into the build process. By default
    it is not enabled, leaving an unchanged version of the normal TLS/HTTP
    client in place. It needs to be explicitly enabled by passing
    --with-libcurl to configure.
    
    A file is added from the libcurl distribution (libcurl.m4) to handle
    detection of the library and the needed parameters. libcurl.m4 is
    licensed under MIT. This should not be a problem.
    
    The majority of other changes are to the
    daemon/connect/WebDownloader.cpp file. This contains the majority of
    code that interacts with libcurl proper. An effort has been made to
    reuse as much of the existing infrastructure, and provide a clean
    separation between "classic" and "libcurl" code.
    
    Small changes ere made to daemon/feed/FeedCoordinator.cpp and
    daemon/queue/UrlCoordinator.cpp to deal with subclasses to the
    WebDownloader class, which might also initiate HTTPS connections. As
    libcurl is a C library, callbacks into C++ code require some shims.
    
    While the WebDownloader class code is tested somewhat, the
    UrlCoordinator and FeedCoordinator code is not, for lack of knowledge as
    to how that code path is triggered.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -83,6 +83,9 @@
 /* Define to 1 if you have the <inttypes.h> header file. */
 #undef HAVE_INTTYPES_H
 
+/* Define to 1 to use libcurl for HTTP downloads. */
+#undef HAVE_LIBCURL
+
 /* Define to 1 to use GnuTLS library for TLS/SSL-support. */
 #undef HAVE_LIBGNUTLS
 
--- a/configure
+++ b/configure
@@ -645,6 +645,9 @@ WITH_PAR2_FALSE
 WITH_PAR2_TRUE
 ncurses_LIBS
 ncurses_CFLAGS
+LIBCURL
+LIBCURL_CPPFLAGS
+_libcurl_config
 libxml2_LIBS
 libxml2_CFLAGS
 PKG_CONFIG_LIBDIR
@@ -764,6 +767,7 @@ enable_largefile
 enable_libxml2
 with_libxml2_includes
 with_libxml2_libraries
+with_libcurl
 enable_curses
 with_libcurses_includes
 with_libcurses_libraries
@@ -1460,6 +1464,8 @@ Optional Packages:
                           libxml2 include directory
   --with-libxml2-libraries=DIR
                           libxml2 library directory
+  --with-libcurl=PREFIX   look for the curl library in PREFIX/lib and headers
+                          in PREFIX/include
   --with-libcurses-includes=DIR
                           libcurses include directory
   --with-libcurses-libraries=DIR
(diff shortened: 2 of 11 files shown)


hugbug:
Thank you for your efforts. However I cannot accept this PR. Adding new dependencies is a no go. I'm making a lot of efforts to ensure the dependency list of the project is as short a possible.

> This is a warning message, and it indicates that the web server is not configured properly

Why not contact the server owners and ask them to fix their issue?

Have you tested how nzbget compiled with OpenSSL behaves? You can use the installer from nzbget home page.

Lalufu:
Thanks for the feedback. I did not attempt to contact the nzbgeek.info folks, I am not a member (the failure was reported to me by someone else), and there's no contact on their start page. There is no shortage of broken web servers, though, TLS is a complex protocol, and so is HTTP.
Building with OpenSSL instead of GnuTLS fixes the issue as well.
I'll probably continue to maintain this for my personal builds.

hugbug:
I believe libcurl uses OpenSSL for TLS. By using libcurl you basically forces the usage of OpenSSL instead of GnuTLS. That alone fixes the issue and not something in libcurl, it seems.

Lalufu:
libcurl can also use both, although this specific version is indeed built against openssl. wget on my fedora is built against gnutls, and also works fine.

The first issue I mentioned in the pull request is in https://github.com/nzbget/nzbget/blob/develop/daemon/connect/TlsSocket.cpp#L343
gnutls_handshake() can return GNUTLS_E_WARNING_ALERT_RECEIVED, which is what happens here, and the code treats it as fatal. As said this is relatively easy to fix (mostly check gnutls_error_is_fatal(), drop a log message, and call gnutls_handshake() again.
For the alert occuring on connection shutdown (Connection not properly terminated) it might be enough to call gnutls_bye() with GNUTLS_SHUT_RDWR, but I haven't tested that.

PR #679, opened 2020-02-10
Fix strncasecmp size parameter off by ones

disconnect3d:
The "http://" string has 7 bytes without nullbyte so the size in `strncasecmp(nzbContent, "http://", 6)` is off by one.

The same goes for `"https://"` string.


hugbug:
Thank you.

Issue #690, opened 2020-04-13
nzbget-nserv: core dump with input "-z 0" 


sanderjo:
I keep on fighting with the input for nzbget's nserv option "-z". I even get core dumps.

I'll send a PR for:
* clearer usage (for people like me): the number is optional, so ... note to self: just leave out any number and leave it to nserv (so: 500000)
* protection against silly input: 0 or negative/overflow numbers

```
$ ./nzbget --nserv -d ~/some-data-dir -z 0 -q
[INFO] NServ 21.1-testing-r2312 (Test NNTP server)
[INFO] Generating nzbs for /home/sander/some-data-dir
Floating point exception (core dumped)
```
```
$ ./nzbget --nserv -d ~/some-data-dir -z something -q
[INFO] NServ 21.1-testing-r2312 (Test NNTP server)
[INFO] Generating nzbs for /home/sander/some-data-dir
Floating point exception (core dumped)
```

resulting in
```
$ ll ~/some-data-dir
total 102408
drwxr-xr-x  2 sander sander      4096 apr 13 11:56 ./
drwxr-xr-x 29 sander sander      4096 apr 13 11:55 ../
-rw-r--r--  1 sander sander 104857600 apr 13 11:55 bla-100M.bin
-rw-r--r--  1 sander sander         0 apr 13 11:56 bla-100M.bin.nzb
```
which will not create a new nzb on the next try with correct input ... 

EDIT

Also unwise usage:
```
./nzbget --nserv -d ~/some-data-dir -z 1 -q
```
which will create a .NZB about 100 times bigger than the file to be handled. So inflicting a DoS on your own system.
I'll build in a protection against that too.

markheloking:
Can this be closed now that the PR is closed? Seems like not many people are using nserv anyway...

PR #691, opened 2020-04-13
nserv -z: show better usage, and protection against wrong input

sanderjo:
Show "-z [seg-size]" with square brackets because seg-size is optional
Override seg-size <= 1000 to default value 500000 to avoid coredump and or self-DoS

See core dump examples on https://github.com/nzbget/nzbget/issues/690

### DIFF
--- a/daemon/nserv/NServMain.cpp
+++ b/daemon/nserv/NServMain.cpp
@@ -162,7 +162,7 @@ void NServPrintUsage(const char* com)
 		"    -v <verbose>    - verbosity level 0..3 (default is 2)\n"
 		"    -w <msec>       - response latency (in milliseconds)\n"
 		"    -r <KB/s>       - speed throttling (in kilobytes per second)\n"
-		"    -z <seg-size>   - generate nzbs for all files in data-dir (size in bytes)\n"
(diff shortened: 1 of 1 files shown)


sanderjo:
EDIT: hmmm, I have check if this patch works as I wanted. This is not how it should be:

```
$ ./nzbget --nserv -d ~/post_this/ -z -q
[INFO] NServ 21.1-testing-r2312M (Test NNTP server)
[INFO] Generating nzbs for /home/sander/post_this/
[INFO] Generating nzb for my-10GB.bin
[DETAIL] Processing my-10GB.bin
[INFO] Nzb generation finished
[INFO] Press Ctrl+C to quit
[INFO] Listening on port 6791
```
... the `-q` is ignored, probably because it is 'eaten' by `-z` as parameter ...

hugbug:
Sorry, can't accept this PR. 
1. All segment sizes should be accepted, even 1 byte. 0 and negative can be indeed checked and reported.
2. Support for default segment size requires more work but you figured that out already.

Considering the nserv is used only by two persons I would not spend any time improving this.

sanderjo:
I agree: I dived into getopts, and it should handle optional parameters, but even did I can't get working: it eats up the next "-<option>". Weird.

PR #739, opened 2021-03-12
fix nserv group

lucasheld:
NServ currently removes the first character from group name. The error is fixed by correcting the offset.


hugbug:
Thank you!

Issue #741, opened 2021-04-08
Crashing every minute


Toucan-Sam:
Hey guys, at the moment I'm getting a crash every minute or so that results in NZBGet reloading.
```
[INFO] nzbget 21.0 server-mode
[INFO] nzbget 21.0 server-mode
[INFO] nzbget 21.0 server-mode
```
This causes everything to stop, but then continues downloading for a minute or so, then restarts again.
I'm using the linuxserver.io container, but posting here because I don't think it's container specific.

Previously (and still occasionally) I get
```
[ERROR] Binding socket failed for 0.0.0.0: ErrNo 98, Address in use
[ERROR] Binding socket failed for 0.0.0.0: ErrNo 98, Address in use
[ERROR] Binding socket failed for 0.0.0.0: ErrNo 98, Address in use
[ERROR] Binding socket failed for 0.0.0.0: ErrNo 98, Address in use
```
But there is nothing else on or trying to be on the same port
```
# netstat -ntlpa | grep 6789
#
```
If anyone has thoughts or suggestions, I'd appreciate them. :)

zombiehoffa:
are you running the linuxserver container in docker? I am and this started happening to me today as well.

Toucan-Sam:
@zombiehoffa Yep. Ubuntu 20.04.

zombiehoffa:
ubuntu 20.04 running as an lxc in proxmox over here. I also tried testing, same error. Nothing has really changed on this system and there are no obvious things binding to that port.

Toucan-Sam:
The errors I'm getting at the moment is
`[36509.026027] nzbget[191497]: segfault at 0 ip 00007f90d89f2352 sp 00007f90d7ff1bb0 error 4 in ld-musl-x86_64.so.1[7f90d89b7000+48000]`

Surely there must be more people with this issue?
Some additional links I've found that might be related:
https://jira.hyperledger.org/browse/BE-743 which links to...
https://github.com/nodejs/docker-node/issues/1158
https://github.com/nodejs/docker-node/issues/1158#issuecomment-577261378

Toucan-Sam:
Ultimately I've been able to solve this issue by clearing things from the intermediate, queue, nzb, and tmp folders in the downloads directory. I don't know exactly what the issue is/was but I assume it's crashing trying to interpret a particular character or long file name.

[...]

PR #745, opened 2021-04-14
Fix segfault caused by malformed nzb files without subject

### LINKED ISSUE #744: Server Crash on potential bad NZB from sonarr/radarr
Hi sorry I've been having issues since upgrading my unraid servers.

My issue:

NZBget server seems to crash when being sent certain NZB from sonarr/radarr

issue is present on: 
https://hub.docker.com/layers/linuxserver/nzbget/latest/images/sha256-755ac7458a0a0272751af4dbaf789490f238568937902f03b1e02714f9468ac8?context=explore

issue is not present on:
https://hub.docker.com/layers/linuxserver/nzbget/v21.0-ls70/images/sha256-c5dccc23bcae9bf83e6f9c6f99459ea62fac869e5bd1d23af6300e9414d84ac2?context=explore

ls70 container shows an error in the sonarr UI rather than crashing the NZBGET server.
The error that shows up in sonarr on ls70 is `name cannot begin with the '6' character, hexadecimal value 0x36. Line 8, position 68`

It looks to me like an error is no longer being handle correctly and as such the server is crashing. 

I found my "fix" 

here
https://forum.openmediavault.org/index.php?thread/38669-communication-error-cannot-establish-connection-to-nzbget/

clearing the queue and nzb folders fix the issue until sonarr tries to send a error causing nzb back to nzbget

I wish I had more time to figure out exactly what commit are part of ls70 vs ls71. For now I've reverted to ls70

if there is already an issue for this uncaught error i apologize. Thank you

throwaway481:
fixes #744 and probably #741 

Adding NZB files without a proper subject leads to uninitialised memory access (read & write) on non-windows systems. This may segfault or just corrupt memory.

@hugbug please consider releasing this soon. I have not examined how exploitable this actually is, but worst case this may lead to remote code execution.

Example nzb:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE nzb PUBLIC "-//newzBin//DTD NZB 1.1//EN" "http://www.newzbin.com/DTD/nzb/nzb-1.1.dtd">
<nzb xmlns="http://www.newzbin.com/DTD/2003/nzb">
<head>
 <meta type="name">Unimportant</meta>
 <meta type="category">Unimportant</meta>
</head>
<file poster="dummy" date="1234567890">
 <groups>
  <group>alt.binaries.unimportant</group>
 </groups>
 <segments>
  <segment bytes="1" number="1">unimportant</segment>
 </segments>
</file>
</nzb>
```

I haven't programmed C/C++ in 15 years, so feel free to replace the fix with something more idiomatic if you want :-)

### DIFF
--- a/daemon/queue/DownloadInfo.h
+++ b/daemon/queue/DownloadInfo.h
@@ -209,9 +209,9 @@ private:
 	ArticleList m_articles;
 	Groups m_groups;
 	ServerStatList m_serverStats;
-	CString m_subject;
-	CString m_filename;
-	CString m_origname;
+	CString m_subject = "";
+	CString m_filename = "";
+	CString m_origname = "";
 	int64 m_size = 0;
 	int64 m_remainingSize = 0;
 	int64 m_successSize = 0;
--- a/daemon/queue/NzbFile.cpp
+++ b/daemon/queue/NzbFile.cpp
@@ -660,6 +660,10 @@ void NzbFile::Parse_StartElement(const char *name, const char **atts)
 				m_fileInfo->SetTime(atoi(attrvalue));
 			}
 		}
+		if (!strcmp("", m_fileInfo->GetSubject())) {
+			m_nzbInfo->AddMessage(Message::mkWarning, "Malformed nzb-file, <file> without attribute \"subject\"");
+			return;
+		}	
 	}
 	else if (!strcmp("segment", name))
 	{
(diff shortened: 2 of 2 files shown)


throwaway481:
Also... I just fixed this single uninitialised memory access that led to a crash with real world nzb data. Other structures may be affected as well, because of uninitialised strings in the structures. Someone may want to check this.

hugbug:
Thank you.
Can you please send me a couple of real files? nzbget@gmail.com.
I'd like to check how to process them best. Obviously the program shouldn't crash. The question is if the file-elements contain any useful data and should be kept. Filename can be automatically generated at this stage. The real filename is read later from the article anyway.

hugbug:
I've fixed this as follows: for file elements not having subject attribute an (almost) random subject is generated and is used as file name; correct file names are read from articles later during downloading.

Tested with the example real nzb-files and they were downloaded successfully without issues.

Thank you for reporting this and for the pull request.

BTW there are no uninitialised memory or uninitialised strings. All strings are initialised with null. It's just that sometimes this special value isn't expected in certain contexts ;) The program will then always crash, which is good (for reproducing) and isn't exploitable (I hope so).

PR #777, opened 2021-08-22
Add registry key so Windows will display version number in programs list

ThioJoe:
I'm not very familiar with Nullsoft / NSIS so I'm not sure if there's a more elegant way to do this, or have it automatically pull the version number from somewhere more central. But this should make it create a "DisplayVersion" string in the registry alongside the others, so Windows will actually show the version number in the Add/Remove Programs list.

This makes the program compatible with package managers/updates like [WinGet](https://devblogs.microsoft.com/commandline/windows-package-manager-1-0/), the new official Windows Package Manager.

### DIFF
--- a/windows/nzbget-setup.nsi
+++ b/windows/nzbget-setup.nsi
@@ -41,6 +41,7 @@
 
 Name "NZBGet"
 OutFile "nzbget-setup.exe"
+Var displayVersion "21.1"
 
 ;Default installation folder
 InstallDir "$PROGRAMFILES64\NZBGet"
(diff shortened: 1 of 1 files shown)


zotabee:
+1

Thank you ThioJoe.

Hopefully it fixes the bug with WinGet trying to update NZBGet each time because it seems it's missing the version number in the registered information for every installed apps on Windows.

zotabee:
As for users playing a lot with winget upgrade and/or winget import, each time it will propose to re-install nzbget. That can be annoying so here you are a temporary fix / workaround thanks to @ThioJoe 

To add the DisplayVersion "21.1" key yourself (for Win11 x64 systems) :

- Create a text file
- Copy/paste this inside the text file and save it:

```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\NZBGet]
"DisplayVersion"="21.1"
```

- Rename the extension to .reg
- Run it by double clicking on the .reg, confirm, done.

Now there will be the version number shown in the control panel and winget won't detect NZBGet with "unknown version" anymore.

Hopefully it will be fixed one day with the next update of NZBGet.

PR #793, opened 2021-12-07
daemon:connect: don't use FIPS_mode_set with OpenSSL 3

schopin-pro:
This function has been removed in OpenSSL 3, replaced by
EVP_default_properties_enable_fips. See
https://www.openssl.org/docs/man3.0/man7/migration_guide.html

### DIFF
--- a/daemon/connect/TlsSocket.cpp
+++ b/daemon/connect/TlsSocket.cpp
@@ -189,7 +189,11 @@ void TlsSocket::Final()
 
 #ifdef HAVE_OPENSSL
 #ifndef LIBRESSL_VERSION_NUMBER
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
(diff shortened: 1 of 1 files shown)


schopin-pro:
FYI, this patch is being shipped in the Ubuntu packaging of nzbget as part of our OpenSSL 3.0 transition.

sbraz:
Hi @hugbug, could you please look into this?

sanderjo:
For Google searchers: solves this nzbget error during nzbget's `make`:

`daemon/connect/TlsSocket.cpp:192:9: error: ‘FIPS_mode_set’ was not declared in this scope`

This is how I applied the patch / pull request (just git basics):

```
git clone https://github.com/nzbget/nzbget.git
cd nzbget/
git pull origin pull/793/head
```
and then the `./configure` and `make` just work.

Thank you, @schopin-pro

PR #832, opened 2022-10-25
unmangle the deliberately-malformed subject from WtF release group

paul-chambers:
There is at least one release group (perhaps more) that deliberately mangles the subject field. The result is files named like '`abc.xyz.`_{hex digits}_' which is hugely unhelpful.

This PR adds some additional logic to identify these malformed headers and 'unmangle' them, so the extracted files are given the correct names. It has no effect in the absence of these intentionally-malformed headers.

I've been using this modified version for a few months now, and it has cured the problem for me.


SteffanCline:
Apologies but I'm a novice. Is there a place I can download a build of this to test and see if it resolves my issues? I have a bunch of these that are completed but I can't seem to do anything with them. I'm hoping with your build I can redo the post processing and correct them all.

